PR #1, opened 2021-04-15
Hw1 task1

E1-Bart0:
Homework1 Task1 Done


All_home_works/hw1/task1.py
```diff
@@ -0,0 +1,35 @@
...
+
+def get_fib_first_element(data, fib_sequence):
+    fib_num = next(fib_sequence)
+    while fib_num < data[0]:
+        fib_num = next(fib_sequence)
+    return fib_num
+
+
+def check_fibonacci(data: Sequence[int]) -> bool:
+    if not len(data):
```
Cjkjvfnby: Why do you need to check `len`?

test/hw1/test_task1.py
```diff
@@ -0,0 +1,165 @@
...
+        987,
+        1597,
+        2584,
+        4181,
+    )
+    assert check_fibonacci(data)
+
+
+def test_check_valid_data__not_from_begin():
+    data = (
```
Cjkjvfnby, line 15: I'm pretty sure that you just copy-pasted these values from the output of your code.

Test that checks that one run of code ie equals another run of the same code only test that your output is stable. It has nothing with checking if the output is correct.

Cjkjvfnby (changes requested):
I don't know if the tests check the correct values. 

Requirement for test: it is easy to check that values are correct only by looking at them.    

To achieve it use small numbers, use a short sequence.  The sequence of 5th and 6th elements behaves pretty same as 5000th and 5001st. 

So let's use only numbers from fib[:7] and sequence with size <=5 for testing.

test/hw1/test_task1.py
```diff
@@ -20,11 +20,6 @@ def test_check_valid_data__not_from_begin():
     assert check_fibonacci(data)
 
 
-def test_check_no_data():
```
Cjkjvfnby: Why did you remove this test?

Cjkjvfnby (changes requested):

test/hw1/test_task1.py
E1-Bart0: Because you asked why I am checking the length of data. So I thought it's unnecessary and removed the test.

E1-Bart0: I see. "If len(data):" is the same way "if data:". Did you want this from me?

Cjkjvfnby: > "if data:"

Yes, that is what I want.

Cjkjvfnby (approved):

PR #2, opened 2021-04-15
Hw1 task2

E1-Bart0:
Homework1 Task2 Done

### DIFF
--- a/All_home_works/hw1/task2.py
+++ b/All_home_works/hw1/task2.py
@@ -13,13 +13,19 @@ from typing import Tuple
 
 def find_maximum_and_minimum(file_name: str) -> Tuple[int, int]:
     with open(file_name) as data_file:
-        min_value = max_value = int(data_file.readline()[:-1])
+        max_value, min_value = get_first_value_in(data_file)
         for line in data_file.readlines():
             line = line[:-1]
             if line:
                 num = int(line)
                 max_value, min_value = check_max_min_value(max_value, min_value, num)
     return min_value, max_value
+
+
+def get_first_value_in(data_file):
+    line = data_file.readline()[:-1]
+    min_value = max_value = int(line)
+    return max_value, min_value
 
 
 def check_max_min_value(max_value, min_value, num):
--- a/All_home_works/hw1/task2.py
+++ b/All_home_works/hw1/task2.py
@@ -1,0 +1,30 @@
+"""
+Write down the function, which reads input line-by-line, and find maximum and minimum values.
+Function should return a tuple with the max and min values.
+For example for [1, 2, 3, 4, 5], function should return [1, 5]
+We guarantee, that file exists and contains line-delimited integers.
+To read file line-by-line you can use this snippet:
+with open("some_file.txt") as fi:
+    for line in fi:
+        ...
+"""
+from typing import Tuple
+
+
+def find_maximum_and_minimum(file_name: str) -> Tuple[int, int]:
+    with open(file_name) as data_file:
+        min_value = max_value = int(data_file.readline()[:-1])
+        for line in data_file.readlines():
+            line = line[:-1]
+            if line:
+                num = int(line)
+                max_value, min_value = check_max_min_value(max_value, min_value, num)
+    return min_value, max_value
+
+
+def check_max_min_value(max_value, min_value, num):
+    if num > max_value:
+        max_value = num
+    elif num < min_value:
+        min_value = num
+    return max_value, min_value
--- a/All_home_works/hw1/task2.py
+++ b/All_home_works/hw1/task2.py
@@ -14,7 +14,7 @@ from typing import Tuple
 def find_maximum_and_minimum(file_name: str) -> Tuple[int, int]:
     with open(file_name) as data_file:
         max_value, min_value = get_first_value_in(data_file)
-        for line in data_file.readlines():
+        for line in data_file:
             line = line[:-1]
             if line:
                 num = int(line)
@@ -24,7 +24,10 @@ def find_maximum_and_minimum(file_name: str) -> Tuple[int, int]:
 
 def get_first_value_in(data_file):
     line = data_file.readline()[:-1]
-    min_value = max_value = int(line)
+    min_value, max_value, = (
+        int(line),
+        int(line),
+    )
     return max_value, min_value
 
 
--- a/All_home_works/hw1/task2.py
+++ b/All_home_works/hw1/task2.py
@@ -23,7 +23,7 @@ def find_maximum_and_minimum(file_name: str) -> Tuple[int, int]:
 
 
 def get_first_value_in(data_file):
-    line = data_file.readline()[:-1]
+    line = data_file.readline().rstrip("\n")
     min_value, max_value, = (
         int(line),
         int(line),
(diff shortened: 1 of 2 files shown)


All_home_works/hw1/task2.py
```diff
@@ -0,0 +1,30 @@
...
+    for line in fi:
+        ...
+"""
+from typing import Tuple
+
+
+def find_maximum_and_minimum(file_name: str) -> Tuple[int, int]:
+    with open(file_name) as data_file:
+        min_value = max_value = int(data_file.readline()[:-1])
+        for line in data_file.readlines():
```
Cjkjvfnby: What does data_file.readlines() does?

All_home_works/hw1/task2.py
```diff
@@ -0,0 +1,30 @@
...
+            line = line[:-1]
```
Cjkjvfnby: This operation does not show your intention.

- please add a test for the case when this code will fail.
- please rewrite this line to show your intent clearly.

test/hw1/test_task2.py
```diff
@@ -0,0 +1,39 @@
+import os
+import unittest
+
+from All_home_works.hw1.task2 import find_maximum_and_minimum
+
+
+class Task3Test(unittest.TestCase):
+    def setUp(self):
+        data = [
```
Cjkjvfnby, line 9: Do you really need so many integers to test your code?

test/hw1/test_task2.py
```diff
@@ -0,0 +1,39 @@
...
+            8,
+            23,
+            5,
+            65,
+            -321,
+        ]
+        self.min_value = min(data)
+        self.max_value = max(data)
+
+        self.file_name = "test_task3.txt"
```
Cjkjvfnby: Why not use temp folder?

Cjkjvfnby (changes requested):

All_home_works/hw1/task2.py
E1-Bart0: method readline() will read only one string in file and will return it
method readlines() will read all strings in file and will retun list => List[str]

Cjkjvfnby: What problems we may have with `readlines()`?

E1-Bart0: If a file is a huge, a lot of memory will be required to save it in a list. Because readlines() will read the file immediately, not line by line as readline()

Cjkjvfnby: Can we avoid it?

Cjkjvfnby (changes requested):
Please make sure that you have addressed all comments before assigning a review.

https://github.com/E1-Bart0/homework-repo/pull/2#discussion_r615235436

All_home_works/hw1/task2.py
E1-Bart0: Are you talking about line = line[:-1]?
I just think about this whole part of the code with min and max also readlines().
line = line.rstrip('\n'). Is it ok?

Cjkjvfnby: > line = line.rstrip('\n'). Is it ok?

yes. 

>  line = line[:-1]

When this code will fail?

test/hw1/test_task2.py
```diff
@@ -0,0 +1,45 @@
...
+        [file.write(str(num) + "\n") for num in data]
+        file.seek(0)
+        return file
+
+    def test_find_maximum_and_minimum(self):
+        result = find_maximum_and_minimum(self.file.name)
+
+        assert (self.min_value, self.max_value) == result
+
+    def test_find_maximum_and_minimum_fail(self):
```
Cjkjvfnby: Why do they fail? Please reference the requirements for that case.

Cjkjvfnby (changes requested):

test/hw1/test_task2.py
```diff
@@ -0,0 +1,45 @@
+import tempfile
+import unittest
+
+import pytest
+
+from All_home_works.hw1.task2 import find_maximum_and_minimum
+
+
+class Task3Test(unittest.TestCase):
```
Cjkjvfnby: What does this name mean?

All_home_works/hw1/task2.py
E1-Bart0: I thought about the length of the line. But:
line = ""
line[:-1]
It's ok.
So maybe if a line is None

test/hw1/test_task2.py
E1-Bart0: This test was deleted.

test/hw1/test_task2.py
E1-Bart0: Renamed to HomeWork1Task2Tests.

All_home_works/hw1/task2.py
Cjkjvfnby: >  So maybe if a line is None

Could deadline() return None?

It is another case.

All_home_works/hw1/task2.py
```diff
@@ -0,0 +1,39 @@
...
+        for line in data_file:
+            line = line.rstrip("\n")
+            if line:
+                num = int(line)
+                max_value, min_value = check_max_min_value(max_value, min_value, num)
+    return min_value, max_value
+
+
+def get_first_value_in(data_file):
+    line = data_file.readline()[:-1]
```
Cjkjvfnby: Why do you need `[:-1]`?

Cjkjvfnby (changes requested):

All_home_works/hw1/task2.py
E1-Bart0: [:-1] -> rstrip('\n')

All_home_works/hw1/task2.py
E1-Bart0: if readline() returns an empty string, the end of the file has been reached.
So I don't know when "str[:-1]" will fail

Cjkjvfnby: `str[:-1]` will not fail, but it's will have a different result than `line.rstrip('\n')` and your application will get incorrect data.

E1-Bart0: Well. Do you have any more questions?

Cjkjvfnby: No, only why using the `str[:-1]` in that context is a bug.

E1-Bart0: Sorry, I don't know. The line always ends with '\n' and the last line is empty.

E1-Bart0: Maybe it will be better to use open('text', new line="") and forget about line cutting. But why using the str[:-1] in that context is a bug?

Cjkjvfnby: >  The line always ends with '\n' and the last line is empty.

So the last line does not come with the `\n`
Can we create a file where the last line is not '\n'?  (Please do experiments)

Cjkjvfnby: > Maybe it will be better to use open('text', new line="") and forget about line cutting. But why using the str[:-1] in that context is a bug?

Can you give me a working code sample?

E1-Bart0: OK. I understand! If the last line is the last, I mean, no last blank line. My code will fail. Because the line does not end with '\n'.
And about newline="". It's not working. 
Thanks for your patience=)

Cjkjvfnby (approved):

PR #3, opened 2021-04-15
Hw1 task3

E1-Bart0:
Homework1 Task3. 
!!! But Changed max-cognitive-complexity to 10 !!!

### DIFF
--- a/All_home_works/hw1/task3.py
+++ b/All_home_works/hw1/task3.py
@@ -4,21 +4,38 @@ Given four lists A, B, C, D of integer values,
     compute how many tuples (i, j, k, l) there are such that A[i] + B[j] + C[k] + D[l] is zero.
 We guarantee, that all A, B, C, D have same length of N where 0 ≤ N ≤ 1000.
 """
+from collections import Counter
 from typing import List
 
 
-def check_sum_of_four(a: List[int], b: List[int], c: List[int], d: List[int]) -> int:
-    result = 0
+def generate_sum_of_all_variant_of_two_lists(list1, list2):
+    for num1 in list1:
+        for num2 in list2:
+            yield num1 + num2
 
-    for num_a, num_b, num_c, num_d in generate_all_variant(a, b, c, d):
-        if num_a + num_b + num_c + num_d == 0:
-            result += 1
+
+def solution_with_n_in_2(a: List[int], b: List[int], c: List[int], d: List[int]) -> int:
+    result = 0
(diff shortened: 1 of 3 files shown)


All_home_works/hw1/task3.py
```diff
@@ -0,0 +1,24 @@
...
+def check_sum_of_four(a: List[int], b: List[int], c: List[int], d: List[int]) -> int:
+    result = 0
+
+    for num_a, num_b, num_c, num_d in generate_all_variant(a, b, c, d):
+        if num_a + num_b + num_c + num_d == 0:
+            result += 1
+    return result
+
+
+def generate_all_variant(a, b, c, d):
```
Cjkjvfnby: What is the algirithmical complexity of this code?

test/hw1/test_task3.py
```diff
@@ -0,0 +1,16 @@
...
+def test_check_sum_of_four():
+    data = [[7, -7, -5], [-7, 5, -2], [-7, 0, -10], [-3, 0, 2]]
+
+    result = check_sum_of_four(*data)
+    assert 4 == result
+
+
+def test_check_sum_of_four_with_zeros():
+    length_list = 10
+    data = [[0] * length_list for _ in range(4)]
```
Cjkjvfnby: Please don't generate test data. Keep it simple.

Cjkjvfnby (changes requested):

All_home_works/hw1/task3.py
E1-Bart0: (n^4)

Cjkjvfnby: Let's think about the n^3 solution.

E1-Bart0: I think a lot about it. It's hardly imaging for me to solve it with n^3. Or maybe we will consume a lot of memory to keep n^3 numbers in it.
Is it possible?

Cjkjvfnby: The expected solution is O(n^2) complexity with O(n^2) extra memory.

But at the first I want you to find the solution with O(n^3) and O(n). They use the same approach it will be easy to go from one to another. 

You need a different data structure to solve this task.

E1-Bart0: Ok. I tried to make it not memory-consuming. But now I got it.

Cjkjvfnby (approved):

PR #4, opened 2021-04-15
Hw1 task4

E1-Bart0:
Homework1 Task4 Done

### DIFF
--- a/All_home_works/hw1/task4.py
+++ b/All_home_works/hw1/task4.py
@@ -10,23 +10,21 @@ from typing import List
 
 
 def find_maximal_subarray_sum(nums: List[int], k: int) -> int:
-    queue = []
-    max_sum = nums[0]
+    counter = 0
+    max_sum = 0
 
-    for num in nums:
-        if len(queue) == k:
-            queue.pop(0)
-        queue.append(num)
-
-        max_sum = get_max_sum(max_sum, queue)
-
+    for index, current_num in enumerate(nums):
+        if index < k:
+            counter += current_num
+            max_sum += current_num
+        else:
+            num_k_indexes_ago = nums[index - k]
+            counter += current_num - num_k_indexes_ago
+            max_sum = _get_max_sum(counter, max_sum)
     return max_sum
 
 
-def get_max_sum(max_sum, queue):
-    queue_copy = queue.copy()
-    while len(queue_copy):
-        if sum(queue_copy) > max_sum:
-            max_sum = sum(queue_copy)
-        queue_copy.pop(0)
-    return max_sum
+def _get_max_sum(counter, max_sum):
+    if counter > max_sum:
+        max_sum = counter
+    return max_sum  # noqa: R504
--- a/All_home_works/hw1/task4.py
+++ b/All_home_works/hw1/task4.py
@@ -1,0 +1,32 @@
+"""
+Given a list of integers numbers "nums".
+You need to find a sub-array with length less equal to "k", with maximal sum.
+The written function should return the sum of this sub-array.
+Examples:
+    nums = [1, 3, -1, -3, 5, 3, 6, 7], k = 3
+    result = 16
+"""
+from typing import List
+
+
+def find_maximal_subarray_sum(nums: List[int], k: int) -> int:
+    queue = []
+    max_sum = nums[0]
+
+    for num in nums:
+        if len(queue) == k:
+            queue.pop(0)
+        queue.append(num)
+
+        max_sum = get_max_sum(max_sum, queue)
+
+    return max_sum
+
+
+def get_max_sum(max_sum, queue):
+    queue_copy = queue.copy()
+    while len(queue_copy):
+        if sum(queue_copy) > max_sum:
+            max_sum = sum(queue_copy)
+        queue_copy.pop(0)
+    return max_sum
--- a/All_home_works/hw1/task4.py
+++ b/All_home_works/hw1/task4.py
@@ -10,21 +10,19 @@ from typing import List
 
 
 def find_maximal_subarray_sum(nums: List[int], k: int) -> int:
-    counter = 0
-    max_sum = 0
-
-    for index, current_num in enumerate(nums):
-        if index < k:
-            counter += current_num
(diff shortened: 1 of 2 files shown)


test/hw1/test_task4.py
```diff
@@ -0,0 +1,202 @@
+from All_home_works.hw1.task4 import find_maximal_subarray_sum
+
+
+def test_find_maximal_subarray_sum__init_data():
+    nums = [1, 3, -1, -3, 5, 3, 6, 7]
+
```
Cjkjvfnby: Why do you add this empty line here?

All_home_works/hw1/task4.py
```diff
@@ -0,0 +1,32 @@
...
+        queue.append(num)
+
+        max_sum = get_max_sum(max_sum, queue)
+
+    return max_sum
+
+
+def get_max_sum(max_sum, queue):
+    queue_copy = queue.copy()
+    while len(queue_copy):
```
Cjkjvfnby: Why len?

All_home_works/hw1/task4.py
```diff
@@ -0,0 +1,32 @@
...
+from typing import List
+
+
+def find_maximal_subarray_sum(nums: List[int], k: int) -> int:
+    queue = []
+    max_sum = nums[0]
+
+    for num in nums:
+        if len(queue) == k:
+            queue.pop(0)
```
Cjkjvfnby: What is the complexity of this line?

Cjkjvfnby (changes requested):
This solution is hard to read and understand.  You create a lot of lists and mutate them.

I'd recommend rewriting it in a different way.  try to make it simple to follow.

Also, I'd recommend working with indexes instead of creating copies. It looks more suitable for this particular task.

All_home_works/hw1/task4.py
E1-Bart0: O(n)

E1-Bart0: But what complexity if queue from collections will be used (It will be linked list. So it means O(1))?

E1-Bart0: Sorry for grammar)

test/hw1/test_task4.py
E1-Bart0: Nevermind, my bad.

All_home_works/hw1/task4.py
Cjkjvfnby: > Sorry for grammar)

Try https://app.grammarly.com/. It has a plugin for the browser.

E1-Bart0: Thanks. It will help a lot.

All_home_works/hw1/task4.py
```diff
@@ -0,0 +1,30 @@
...
+            counter += current_num
+            max_sum += current_num
+        else:
+            num_k_indexes_ago = nums[index - k]
+            counter += current_num - num_k_indexes_ago
+            max_sum = _get_max_sum(counter, max_sum)
+    return max_sum
+
+
+def _get_max_sum(counter, max_sum):
```
Cjkjvfnby: This function looks familiar, is anything similar in the standard library?

All_home_works/hw1/task4.py
```diff
@@ -0,0 +1,30 @@
+"""
+Given a list of integers numbers "nums".
+You need to find a sub-array with length less equal to "k", with maximal sum.
```
Cjkjvfnby, line 3: What does `less equal to "k"` mean?

Cjkjvfnby (changes requested):

All_home_works/hw1/task4.py
E1-Bart0: I got it. I can suggest solving this, but with complexity (n*k), is it ok?

All_home_works/hw1/task4.py
E1-Bart0: In the standard library, we have both: sum() and max(). Maybe it's not a very good naming of my function, and I do not see a lot of sense do it as a function, but flake8 has a problem with cognitive complexity. It's 7 or 8 if I do not separate it into functions.

Cjkjvfnby: What this function has common to sum() and max() besides the name?

All_home_works/hw1/task4.py
Cjkjvfnby: Yes, it's ok.

test/hw1/test_task4.py
```diff
@@ -0,0 +1,43 @@
+from All_home_works.hw1.task4 import find_maximal_subarray_sum
+
+
+def test_find_maximal_subarray_sum__init_data__end_subarray():
```
Cjkjvfnby, line 4: Actually in that case you can remove `find_maximal_subarray_sum` from the test name because you have this information in the file.

test/hw1/test_task4.py
```diff
@@ -0,0 +1,43 @@
...
+    nums = [1, 3, -1, -3, 5, 3, 6, 7]
+    k = 3
+    result = 16
+    assert result == find_maximal_subarray_sum(nums, k)
+
+
+def test_find_maximal_subarray_sum__middle_subarray():
```
Cjkjvfnby, line 11: You definitely have your own naming convention for that tests.  This approach might be used when you work closely with the team and everyone are on the same page.  In that case, you can try to use some common English phrases, so readers can understand them without additional context. Just adding a couple of new words to the test name make it easier to understand. 

You can try to build an English phrase and then put it to a test name.

- test_find_maximal_subarray_sum__subarray_in_the_middle
- test_subarray_in_the_middle

Cjkjvfnby (approved):

All_home_works/hw1/task4.py
E1-Bart0: Now, this function is another. Sorry, but I still do not understand it.

Cjkjvfnby: The old version could be replaced with `max`. You just implemented something that already present in the standard library.

PR #6, opened 2021-04-15
Hw2 task1

E1-Bart0:
HomeWork2 Task1 Done

### DIFF
--- a/All_home_works/hw2/task1.py
+++ b/All_home_works/hw2/task1.py
@@ -9,20 +9,22 @@ Given a file containing text. Complete using only default collections:
 import string
 from typing import List
 
-CROP_PUNCTUATION = str.maketrans({char: "" for char in string.punctuation})
+PUNCTUATION = {char: "" for char in string.punctuation}
+CROP_PUNCTUATION = str.maketrans(PUNCTUATION)
 
 
 class Counter:
     def __init__(self):
         self.counter = {}
 
-    def count(self, key):
+    def add(self, key):
         if key in self.counter:
             self.counter[key] += 1
         else:
             self.counter[key] = 0
 
-    def most_common(self):
+    @property
+    def sort_counter(self):
         sorted_counter = sorted(
             self.counter.items(), key=lambda item: item[1], reverse=True
         )
@@ -44,7 +46,7 @@ def get_longest_diverse_words(file_path: str) -> List[str]:
         words = [(word, len(set(word))) for word in line.split() if word]
         queue = _check_words(queue, words)
 
-    return [word[0] for word in reversed(queue)]
+    return [word[0] for word in sorted(queue, key=lambda item: item[1], reverse=True)]
 
 
 def _check_words(queue, words):
@@ -53,53 +55,35 @@ def _check_words(queue, words):
             queue.append((word, len_word))
             continue
 
-        queue = sorted(queue, key=lambda x: x[1])
-        if len_word > queue[0][1]:
-            queue.append((word, len_word))
-            queue.pop(0)
+        index_word_len = min(enumerate(queue), key=lambda item: item[1][1])
+        if len_word > index_word_len[1][1]:
+            queue[index_word_len[0]] = (word, len_word)
     return queue
 
 
 def get_rarest_char(file_path: str) -> str:
     counter = Counter()
     for line in _get_lines(file_path):
-        _check_char_in(line, counter)
-    return list(reversed(counter.most_common()))[0][0]
-
-
-def _check_char_in(line, counter):
-    for char in line:
-        if char not in string.punctuation and ord(char) < 126:
-            counter.count(char)
+        [counter.add(char) for char in line if char.isalpha()]
+    return list(reversed(counter.sort_counter))[0][0]
 
 
 def count_punctuation_chars(file_path: str) -> int:
     counter = 0
     for line in _get_lines(file_path):
-        for char in line:
-            counter = _check_punctuation(char, counter)
+        counter += sum([1 for char in line if char in PUNCTUATION])
     return counter
 
 
-def _check_punctuation(char, counter):
-    if char in string.punctuation:
-        counter += 1
-    return counter
-
-
-def count_non_ascii_chars(file_path: str) -> int:  # noqa: CCR001
+def count_non_ascii_chars(file_path: str) -> int:
     counter = 0
     for line in _get_lines(file_path):
-        for char in line:
-            if ord(char) > 126:
-                counter += 1
+        counter += sum([1 for char in line if not char.isascii()])
     return counter
 
 
-def get_most_common_non_ascii_char(file_path: str) -> str:  # noqa: CCR001
+def get_most_common_non_ascii_char(file_path: str) -> str:
     counter = Counter()
     for line in _get_lines(file_path):
-        for char in line:
-            if ord(char) > 126:
-                counter.count(char)
-    return counter.most_common()[0][0]
+        [counter.add(char) for char in line if not char.isascii()]
+    return counter.sort_counter[0][0]
--- a/All_home_works/hw2/task1.py
+++ b/All_home_works/hw2/task1.py
@@ -10,13 +10,13 @@ import string
 from collections import Counter, deque
 from typing import List
 
-TRANSLATE = str.maketrans({char: "" for char in string.punctuation})
+CROP_PUNCTUATION = str.maketrans({char: "" for char in string.punctuation})
 
 
 def _get_lines(file_path: str) -> str:
     with open(file_path, encoding="raw_unicode_escape") as reading_file:
-        for line in reading_file.readlines():
-            line = line[:-1]
+        for line in reading_file:
+            line = line.rstrip("\n")
             if line:
                 yield line
 
@@ -24,7 +24,7 @@ def _get_lines(file_path: str) -> str:
 def get_longest_diverse_words(file_path: str) -> List[str]:
     queue = deque()
     for line in _get_lines(file_path):
-        line = line.translate(TRANSLATE)
+        line = line.translate(CROP_PUNCTUATION)
         words = [(word, len(set(word))) for word in line.split() if word]
         queue = _check_words(queue, words)
 
--- a/All_home_works/hw2/task1.py
+++ b/All_home_works/hw2/task1.py
@@ -7,10 +7,26 @@ Given a file containing text. Complete using only default collections:
     5) Find most common non ascii char for document
 """
 import string
-from collections import Counter, deque
 from typing import List
 
 CROP_PUNCTUATION = str.maketrans({char: "" for char in string.punctuation})
+
+
+class Counter:
+    def __init__(self):
+        self.counter = {}
+
+    def count(self, key):
+        if key in self.counter:
+            self.counter[key] += 1
+        else:
+            self.counter[key] = 0
+
+    def most_common(self):
+        sorted_counter = sorted(
+            self.counter.items(), key=lambda item: item[1], reverse=True
+        )
+        return list(sorted_counter)
 
 
 def _get_lines(file_path: str) -> str:
@@ -22,7 +38,7 @@ def _get_lines(file_path: str) -> str:
 
 
 def get_longest_diverse_words(file_path: str) -> List[str]:
-    queue = deque()
+    queue = []
     for line in _get_lines(file_path):
         line = line.translate(CROP_PUNCTUATION)
         words = [(word, len(set(word))) for word in line.split() if word]
@@ -54,7 +70,7 @@ def get_rarest_char(file_path: str) -> str:
 def _check_char_in(line, counter):
     for char in line:
         if char not in string.punctuation and ord(char) < 126:
-            counter[char] += 1
+            counter.count(char)
 
 
 def count_punctuation_chars(file_path: str) -> int:
@@ -84,7 +100,7 @@ def _check_non_ascii(char, counter):
         if isinstance(counter, int):
             counter += 1
         else:
-            counter[char] += 1
+            counter.count(char)
     return counter
 
 
--- a/All_home_works/hw2/task1.py
+++ b/All_home_works/hw2/task1.py
@@ -1,0 +1,96 @@
+"""
+Given a file containing text. Complete using only default collections:
+    1) Find 10 longest words consisting from largest amount of unique symbols
+    2) Find rarest symbol for document
+    3) Count every punctuation char
+    4) Count every non ascii char
+    5) Find most common non ascii char for document
+"""
+import string
+from collections import Counter, deque
+from typing import List
+
+TRANSLATE = str.maketrans({char: "" for char in string.punctuation})
+
+
+def _get_lines(file_path: str) -> str:
+    with open(file_path, encoding="raw_unicode_escape") as reading_file:
+        for line in reading_file.readlines():
+            line = line[:-1]
+            if line:
+                yield line
+
+
+def get_longest_diverse_words(file_path: str) -> List[str]:
+    queue = deque()
+    for line in _get_lines(file_path):
+        line = line.translate(TRANSLATE)
+        words = [(word, len(set(word))) for word in line.split() if word]
+        queue = _check_words(queue, words)
+
+    return [word[0] for word in reversed(queue)]
+
+
+def _check_words(queue, words):
+    for word, len_word in words:
+        if len(queue) < 10:
+            queue.append((word, len_word))
+            continue
+
+        queue = sorted(queue, key=lambda x: x[1])
+        if len_word > queue[0][1]:
+            queue.append((word, len_word))
+            queue.pop(0)
+    return queue
+
+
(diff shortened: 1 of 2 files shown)


All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,96 @@
...
+import string
+from collections import Counter, deque
+from typing import List
+
+TRANSLATE = str.maketrans({char: "" for char in string.punctuation})
+
+
+def _get_lines(file_path: str) -> str:
+    with open(file_path, encoding="raw_unicode_escape") as reading_file:
+        for line in reading_file.readlines():
```
Cjkjvfnby: We already have a discussion about `readlines()` in the different PR. Please check it.

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,96 @@
...
+            line = line[:-1]
```
Cjkjvfnby: We already have a discussion about such code in the different PR. Please check it.

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,96 @@
...
+    2) Find rarest symbol for document
+    3) Count every punctuation char
+    4) Count every non ascii char
+    5) Find most common non ascii char for document
+"""
+import string
+from collections import Counter, deque
+from typing import List
+
+TRANSLATE = str.maketrans({char: "" for char in string.punctuation})
```
Cjkjvfnby: Can we give a better name for this variable?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,96 @@
...
+    return [word[0] for word in reversed(queue)]
+
+
+def _check_words(queue, words):
+    for word, len_word in words:
+        if len(queue) < 10:
+            queue.append((word, len_word))
+            continue
+
+        queue = sorted(queue, key=lambda x: x[1])
```
Cjkjvfnby: What is the algorithmic and memory complexity of this line?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,96 @@
+"""
+Given a file containing text. Complete using only default collections:
```
Cjkjvfnby, line 2: > Complete using only default collections:

What does it mean?

Cjkjvfnby (changes requested):
I have checked a part of the code. Please fix the comments and re-request a review.

All_home_works/hw2/task1.py
E1-Bart0: I completely don't understand what does it mean. Because Counter and Deque are being imported from collections, it's a built-in library. So we can understand this phrase ("Complete using only default collections:") both: Do not use it at all and use only it. Nonsense.

All_home_works/hw2/task1.py
E1-Bart0: Sorted in python is a TimSort, so that means n*log(n)

All_home_works/hw2/task1.py
Cjkjvfnby: You need to take a lecture into the account, is about builtin collections. In that case, default is means builtin.
The idea of this task is to practice low-level operation.

All_home_works/hw2/task1.py
Cjkjvfnby: What are you trying to achieve with sorting? What is the best complexity for that action?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(self):
+        self.counter = {}
+
+    def count(self, key):
+        if key in self.counter:
+            self.counter[key] += 1
+        else:
+            self.counter[key] = 0
+
+    def most_common(self):
```
Cjkjvfnby, line 31: What is the use case for this method?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,112 @@
...
+from typing import List
+
+CROP_PUNCTUATION = str.maketrans({char: "" for char in string.punctuation})
+
+
+class Counter:
+    def __init__(self):
+        self.counter = {}
+
+    def count(self, key):
```
Cjkjvfnby: What does this function do?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def get_longest_diverse_words(file_path: str) -> List[str]:
+    queue = []
+    for line in _get_lines(file_path):
+        line = line.translate(CROP_PUNCTUATION)
+        words = [(word, len(set(word))) for word in line.split() if word]
+        queue = _check_words(queue, words)
+
+    return [word[0] for word in reversed(queue)]
```
Cjkjvfnby: Why do you need to reverse it, if you could maintain it in proper order?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,112 @@
...
+def get_rarest_char(file_path: str) -> str:
+    counter = Counter()
+    for line in _get_lines(file_path):
+        _check_char_in(line, counter)
+    return list(reversed(counter.most_common()))[0][0]
+
+
+def _check_char_in(line, counter):
+    for char in line:
+        if char not in string.punctuation and ord(char) < 126:
```
Cjkjvfnby: Could you share more information about what happened in this line?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,112 @@
...
+def count_punctuation_chars(file_path: str) -> int:
+    counter = 0
+    for line in _get_lines(file_path):
+        for char in line:
+            counter = _check_punctuation(char, counter)
+    return counter
+
+
+def _check_punctuation(char, counter):
+    if char in string.punctuation:
```
Cjkjvfnby: Can we make it better in terms of O(n)?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,112 @@
...
+def count_non_ascii_chars(file_path: str) -> int:
+    counter = 0
+    for line in _get_lines(file_path):
+        for char in line:
+            counter = _check_non_ascii(char, counter)
+    return counter
+
+
+def _check_non_ascii(char, counter):
+    if ord(char) > 126:
```
Cjkjvfnby: What is an easies way to do it?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,112 @@
...
+        if isinstance(counter, int):
+            counter += 1
+        else:
+            counter.count(char)
```
Cjkjvfnby: Let's avoid instance check, if you need to use this function in a different context write 2 functions.

Or maybe you can change the function to avoid using counter argument?

Cjkjvfnby (changes requested):

All_home_works/hw2/task1.py
E1-Bart0: Honestly speaking, I do not see any sense in that function, because it's too easy. I will make this action in the main function. I think it's even more comfortable to read. But flake8 and his cognitive complexity with 5))

Cjkjvfnby: There is a lot of other ways to get rid of complexity.  

Why do you need a line if you iterate over chars?   This is a place where you can make your code simpler.

```
    for line in _get_lines(file_path):
        for char in line:
```


> `ord(char) > 126:` 

You have 3 tokens in the code.  When you will address my comment, you will have only 2. https://github.com/E1-Bart0/homework-repo/pull/6#discussion_r618680097

All_home_works/hw2/task1.py
Cjkjvfnby: This comment marked as outdated not because you react on it, but because you moved it to a different place.

Cjkjvfnby (changes requested):
Please check that all comments are resolved before assigning me as a reviewer.

All_home_works/hw2/task1.py
E1-Bart0: if we will check it in a dict. It will be O(1)

All_home_works/hw2/task1.py
E1-Bart0: if chat.isalpha():
   ...
?

All_home_works/hw2/task1.py
E1-Bart0: it can be @property. Or I did not understand

All_home_works/hw2/task1.py
E1-Bart0: RENAME: count -> add

All_home_works/hw2/task1.py
E1-Bart0: if not char.isascii():
 ....
?

All_home_works/hw2/task1.py
E1-Bart0: I remade it. I tried not to keep all words in memory, so decided to keep only 10 words in memory. So now it works without sorting and complexity is O(n*k), where k is 10 (constant).

All_home_works/hw2/task1.py
Cjkjvfnby: Dict is close, but we can do better.

All_home_works/hw2/task1.py
Cjkjvfnby: Yes.

All_home_works/hw2/task1.py
E1-Bart0: Well, it means set()=) But it is almost the same structure as a dict(). The advantage of set() is not to keep the value in memory, isn't it?

Cjkjvfnby: > But it is almost the same structure as a dict(). 

Then you have two options, both of them are easy to use, and one is slightly better than the other.   Why you should choose the one which is worse?

> The advantage of set() is not to keep the value in memory, isn't it?

There are 5 basic collections in Python.  All of them serve different purposes. I'll recommend studying them all.
Questions about them in one way or another will be present in the job interviews.  Good news that this knowledge is universal and can be applayed to any programming languages.

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,89 @@
...
+    for line in _get_lines(file_path):
+        line = line.translate(CROP_PUNCTUATION)
+        words = [(word, len(set(word))) for word in line.split() if word]
+        queue = _check_words(queue, words)
+
+    return [word[0] for word in sorted(queue, key=lambda item: item[1], reverse=True)]
+
+
+def _check_words(queue, words):
+    for word, len_word in words:
```
Cjkjvfnby, line 54: Check the documentation about `heapq` and `bisect` modules, they really could help with such tasks.  No need to change the existing code.

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,89 @@
...
+        if len_word > index_word_len[1][1]:
+            queue[index_word_len[0]] = (word, len_word)
+    return queue
+
+
+def get_rarest_char(file_path: str) -> str:
+    counter = Counter()
+    for line in _get_lines(file_path):
+        [counter.add(char) for char in line if char.isalpha()]
+    return list(reversed(counter.sort_counter))[0][0]
```
Cjkjvfnby: What is the complexity of this line?  Can we do better?

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def count_punctuation_chars(file_path: str) -> int:
+    counter = 0
+    for line in _get_lines(file_path):
+        counter += sum([1 for char in line if char in PUNCTUATION])
```
Cjkjvfnby: You already familiar with generators, so you can use generator expressions here.  

```suggestion
        counter += sum(1 for char in line if char in PUNCTUATION)
```

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,89 @@
...
+    counter = 0
+    for line in _get_lines(file_path):
+        counter += sum([1 for char in line if not char.isascii()])
+    return counter
+
+
+def get_most_common_non_ascii_char(file_path: str) -> str:
+    counter = Counter()
+    for line in _get_lines(file_path):
+        [counter.add(char) for char in line if not char.isascii()]
```
Cjkjvfnby: List comprehension without using its result looks strange. You create a list and immediately throw it away.  Let's replace it with the for loop.

Cjkjvfnby (changes requested):

All_home_works/hw2/task1.py
```diff
@@ -0,0 +1,89 @@
...
+    2) Find rarest symbol for document
+    3) Count every punctuation char
+    4) Count every non ascii char
+    5) Find most common non ascii char for document
+"""
+import string
+from typing import List
+
+PUNCTUATION = set(string.punctuation)
+CROP_PUNCTUATION = str.maketrans({char: "" for char in PUNCTUATION})
```
Cjkjvfnby: `str.maketrans` can handle a different number of arguments.  Let's use it in another way.

Cjkjvfnby (changes requested):

All_home_works/hw2/task1.py
E1-Bart0: Then you have two options, both of them are easy to use, and one is slightly better than the other. Why you should choose the one which is worse?

Of course, but I already have this structure (dict structure) in memory. So that's why this question appeared.

All_home_works/hw2/task1.py
E1-Bart0: OK. Thanks)

All_home_works/hw2/task1.py
E1-Bart0: I have already answered this question in the next homework. So it can be done for O(n)

"If the most common element is needed, we can do it by O(n). But if we want to sort all elements n*log(n) needed"

All_home_works/hw2/task1.py
E1-Bart0: CROP_PUNCTUATION = str.maketrans('', '', string.punctuation)

All_home_works/hw2/task1.py
Cjkjvfnby: `Set` and `dict` usually have common implementation. But `set` is slightly better for that case because it does not store values.

All_home_works/hw2/task1.py
Cjkjvfnby: > I have already answered this question in the next homework.

I am not sure about timing, but when you fix something, I expect that you check and fix other similar cases as well.

test/hw2/test_task1.py
```diff
@@ -43,7 +43,7 @@ def test_get_longest_diverse_words(self):
         assert result_start_with == checking[: len(result_start_with)]
 
     def test_get_rarest_char(self):
-        rarest_char = "z"
+        rarest_char = "B"
```
Cjkjvfnby, line 46: If you changed the test and do not change data, that means that the behavior of the code changed.

What was changed in that commit?

Cjkjvfnby (changes requested):

All_home_works/hw2/task1.py
```diff
@@ -30,6 +30,14 @@ def sort_counter(self):
         )
         return list(sorted_counter)
 
+    @property
+    def rarest_element(self):
+        return min(self.sort_counter, key=lambda item: item[1])[0]
```
E1-Bart0: That's why the test changed

Cjkjvfnby: I have already switched my context.  This is just a line for me, can you give me a short description of what was before and why you have changed the result of that function execution. Was it an issue of something else?

E1-Bart0: What is the complexity of this line? Can we do better?
@E1-Bart0 E1-Bart0 2 days ago Author Owner

I remade for O(n). And because there are some rarest chars, new method gives us different from the previous method char

Cjkjvfnby: > gives us different from the previous method char

Why does it give a different result?

E1-Bart0: image

Cjkjvfnby: Can you remove reverse without breaking another code?

Cjkjvfnby: What is the complexity of this line? Can we do better?

Cjkjvfnby (changes requested):

All_home_works/hw2/task1.py
E1-Bart0: Now I'm using the min function(), not sorted. So the complexity is O(n) and we can't do better in this structure.
Maybe a little bit better to use: min(self.counter, key=self.counter.get)

E1-Bart0:
https://github.com/E1-Bart0/homework-repo/pull/6#discussion_r625955268

I made mistake. I wanted to write self.counter instead of self.sort_counter. If you are talking about it. And it's not needed at all to use sort_counter.

Cjkjvfnby (approved):

PR #7, opened 2021-04-15
Hw2 task2

E1-Bart0:
HomeWork2 Task2 Done

### DIFF
--- a/All_home_works/hw2/task2.py
+++ b/All_home_works/hw2/task2.py
@@ -22,10 +22,17 @@ from typing import List, Tuple
 
 
 def major_and_minor_elem(inp: List) -> Tuple[int, int]:
-    count = Counter()
+    counter = Counter()
     for num in inp:
-        count[num] += 1
-    counter = count.most_common()
-    max_common = counter[0][0]
-    min_common = counter[-1][0]
-    return max_common, min_common
+        counter[num] += 1
+    return get_minor_major_from(counter)
+
+
+def get_minor_major_from(counter):
+    minor = major = counter.popitem()
+    for word, counts in counter.items():
+        if counts > major[1]:
+            major = (word, counts)
+        elif counts < minor[1]:
+            minor = (word, counts)
+    return major[0], minor[0]
--- a/All_home_works/hw2/task2.py
+++ b/All_home_works/hw2/task2.py
@@ -1,0 +1,31 @@
+"""
+Given an array of size n, find the most common and the least common elements.
+The most common element is the element that appears more than n // 2 times.
+The least common element is the element that appears fewer than other.
+
+You may assume that the array is non-empty and the most common element
+always exist in the array.
+
+Example 1:
+
+Input: [3,2,3]
+Output: 3, 2
+
+Example 2:
+
+Input: [2,2,1,1,1,2,2]
+Output: 2, 1
+
+"""
+from collections import Counter
+from typing import List, Tuple
+
+
+def major_and_minor_elem(inp: List) -> Tuple[int, int]:
+    count = Counter()
+    for num in inp:
+        count[num] += 1
+    counter = count.most_common()
+    max_common = counter[0][0] if counter[0][1] > len(inp) // 2 else None
+    min_common = counter[-1][0] if counter[-2][1] != counter[-1][1] else None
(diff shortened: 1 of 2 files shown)


All_home_works/hw2/task2.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+from collections import Counter
+from typing import List, Tuple
+
+
+def major_and_minor_elem(inp: List) -> Tuple[int, int]:
+    count = Counter()
+    for num in inp:
+        count[num] += 1
+    counter = count.most_common()
```
Cjkjvfnby: What is the complexity of this line?

test/hw2/test_task2.py
```diff
@@ -0,0 +1,23 @@
...
+    assert output1 == major_and_minor_elem(input1)
+
+    input2 = [2, 2, 1, 1, 1, 2, 2]
+    output2 = 2, 1
+    assert output2 == major_and_minor_elem(input2)
+
+
+def test_major_and_minor_elem_non_most_common():
+    data = [1, 1, 1, 2, 2, 2, 0, 0, 0, 3]
+    output = (None, 3)
```
Cjkjvfnby: Is this case mentioned in the requirements?

Cjkjvfnby (changes requested):

All_home_works/hw2/task2.py
E1-Bart0: Get value from a dictionary with a key consume O(1). But to get the most_common element we should check all values. So it's mean O(n).

test/hw2/test_task2.py
E1-Bart0: In description: "The most common element is the element that appears more than n // 2 times.
The least common element is the element that appears fewer than others."
I decided to check it.

All_home_works/hw2/task2.py
Cjkjvfnby: Keep in mind that this is not an exam. You don't need to remember or calculate the answer.  You just need to find it. 

The answer is in the source code.  You can CTRL + click in IDEA or find it on GitHub in the Python repository.

test/hw2/test_task2.py
Cjkjvfnby: Is anything else is related to it?

E1-Bart0: There is nothing else related to it. Are you trying to say to me that I did unnecessary stuff?
But why they saying this: "more than n // 2 times." From this, I understand that in some data there is no most common element, for example: [1,2,3,4,5,6].

Cjkjvfnby: > Are you trying to say to me that I did unnecessary stuff?

Yes. This is exactly what I am doing.  I try to make you notice some little details in the requirements, namely that phrase: "You may assume that the array is non-empty and the most common element
always exist in the array."

E1-Bart0: Sorry! Haven't notice.

Cjkjvfnby (changes requested):
Please check the comment https://github.com/E1-Bart0/homework-repo/pull/7/files#r615309089

All_home_works/hw2/task2.py
E1-Bart0: n*log(n). Because they use sorted() in it.

Cjkjvfnby: Can we do it in O(n)?

Cjkjvfnby (changes requested):

All_home_works/hw2/task2.py
E1-Bart0: If the most common element is needed, we can do it by O(n). But if we want to sort all elements n*log(n) needed

test/hw2/test_task2.py
```diff
@@ -0,0 +1,13 @@
+from All_home_works.hw2.task2 import major_and_minor_elem
+
+
+def test_major_and_minor_elem_example1_from_description():
```
Cjkjvfnby, line 4: `example1`,  `example2` means that this tests test different cases.  So you need to put this difference into the names.  

Or you may not find any, because these tests actually do the same and one of them could be removed.

Cjkjvfnby (approved):

PR #8, opened 2021-04-15
Hw2 task3

E1-Bart0:
HomeWork2 Task3 Done

### DIFF
--- a/All_home_works/hw2/task3.py
+++ b/All_home_works/hw2/task3.py
@@ -18,18 +18,18 @@ from typing import Any, List
 
 
 def combinations(*args: List[Any]) -> List[List]:
-    res, nums = [], []
+    accumulator = []
     if args:
-        recursion(res, nums, args)
-    return res
+        recursion(accumulator, args)
+    return accumulator
 
 
-def recursion(res, nums, args, index=0):
+def recursion(accumulator, args, index=0, nums=None):  # noqa: CCR001
+    nums = nums or []
     for num in args[index]:
         nums.append(num)
         if len(args) == index + 1:
-            res.append(nums.copy())
+            accumulator.append(nums.copy())
         else:
-            recursion(res, nums, args, index=index + 1)
+            recursion(accumulator, args, index=index + 1, nums=nums)
         nums.pop()
-    return res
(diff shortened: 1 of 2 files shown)


All_home_works/hw2/task3.py
```diff
@@ -0,0 +1,35 @@
...
+
+def recursion(res, nums, args, index=0):
+    for num in args[index]:
+        nums.append(num)
+        if len(args) == index + 1:
+            res.append(nums.copy())
+        else:
+            recursion(res, nums, args, index=index + 1)
+        nums.pop()
+    return res
```
Cjkjvfnby: Why do you need to return it?

All_home_works/hw2/task3.py
```diff
@@ -0,0 +1,35 @@
...
+    [1, 4],
+    [2, 3],
+    [2, 4],
+]
+"""
+from typing import Any, List
+
+
+def combinations(*args: List[Any]) -> List[List]:
+    res, nums = [], []
```
Cjkjvfnby: `res` is usually named `accumulator` in the recursion context.

All_home_works/hw2/task3.py
```diff
@@ -0,0 +1,35 @@
...
+    if args:
+        recursion(res, nums, args)
+    return res
+
+
+def recursion(res, nums, args, index=0):
```
Cjkjvfnby: Could we structure this program in a way when you avoid passing `nums` as an argument?

Cjkjvfnby (changes requested):
What is the limitation for the recursion in Python?

All_home_works/hw2/task3.py
E1-Bart0: As the lecturer said, a limitation for the recursion is something near 3000.
Of course, not needed to return a result. I cut it.

All_home_works/hw2/task3.py
E1-Bart0: No problem!

All_home_works/hw2/task3.py
Cjkjvfnby: Please lunch the console and check what is the limit on your Python.

E1-Bart0: 987

Cjkjvfnby (approved):

PR #9, opened 2021-04-15
Hw2 task4

E1-Bart0:
HomeWork2 Task4 Done

### DIFF
--- a/All_home_works/hw2/task4.py
+++ b/All_home_works/hw2/task4.py
@@ -25,13 +25,12 @@ def cache(func: Callable) -> Callable:
     memo = {}
 
     def wrapper(*args, **kwargs):
-        call = str(args) + str(kwargs)
+        call = f"{args}{kwargs}"
         if call in memo:
             print("Called from Cache")  # noqa: T001
-            return memo[call]
         else:
             print("Evaluate function")  # noqa: T001
             memo[call] = func(*args, **kwargs)
-            return memo[call]
+        return memo[call]
 
     return wrapper
--- a/All_home_works/hw2/task4.py
+++ b/All_home_works/hw2/task4.py
@@ -25,7 +25,7 @@ def cache(func: Callable) -> Callable:
     memo = {}
 
     def wrapper(*args, **kwargs):
-        call = f"{args}{kwargs}"
+        call = (args, tuple(kwargs))
         if call in memo:
             print("Called from Cache")  # noqa: T001
         else:
--- a/All_home_works/hw2/task4.py
+++ b/All_home_works/hw2/task4.py
@@ -27,8 +27,10 @@ def cache(func: Callable) -> Callable:
     def wrapper(*args, **kwargs):
         call = str(args) + str(kwargs)
         if call in memo:
+            print("Called from Cache")  # noqa: T001
             return memo[call]
         else:
+            print("Evaluate function")  # noqa: T001
             memo[call] = func(*args, **kwargs)
             return memo[call]
 
--- a/All_home_works/hw2/task4.py
+++ b/All_home_works/hw2/task4.py
@@ -1,0 +1,35 @@
+"""
+Write a function that accepts another function as an argument. Then it
+should return such a function, so the every call to initial one
+should be cached.
+
+
+def func(a, b):
+    return (a ** b) ** 2
+
+
+cache_func = cache(func)
+
+some = 100, 200
+
+val_1 = cache_func(*some)
+val_2 = cache_func(*some)
+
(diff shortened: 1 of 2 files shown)


All_home_works/hw2/task4.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    memo = {}
+
+    def wrapper(*args, **kwargs):
+        call = str(args) + str(kwargs)
+        if call in memo:
```
Cjkjvfnby, line 29: You can reorder your code to avoid diplications.

All_home_works/hw2/task4.py
```diff
@@ -0,0 +1,35 @@
...
+
+"""
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    memo = {}
+
+    def wrapper(*args, **kwargs):
+        call = str(args) + str(kwargs)
```
Cjkjvfnby: Why do you need it?

test/hw2/test_task4.py
```diff
@@ -0,0 +1,14 @@
+from All_home_works.hw2.task4 import cache
+
+
+def test_cache():
+    some = 100, 200
+
+    cache_fun = cache(func)
+    val1 = cache_fun(*some)
+    val2 = cache_fun(*some)
+    assert val1 == val2
```
Cjkjvfnby: You called the same function twice and got the same result.  I'm not sure if it is cached.

Cjkjvfnby (changes requested):

All_home_works/hw2/task4.py
E1-Bart0: Because immutable keys are needed in a dictionary. And str is immutable.

Cjkjvfnby: I am ok with this limitation if you will name the case when this approach will provide the incorrect result.

test/hw2/test_task4.py
Cjkjvfnby: Each time you change something ask yourself, will test failed if you replace your function with

```python
def cache(fun):
    return fun
```

If test is not failed after that it does not test caching.

Cjkjvfnby (changes requested):

All_home_works/hw2/task4.py
E1-Bart0: You are talking about: "call = of'{args}{kwargs}'" str duplications?
Also, I fixed duplications in "return memo[call]"

All_home_works/hw2/task4.py
E1-Bart0: At first, we will call the function with args, then we will call with kwargs. Example:
   cache(a=1) it will calculate  this;
   cache(1) it will calculate this also, and not take from the cache.

test/hw2/test_task4.py
E1-Bart0: Oh. If I understood clearly, this test will fix it:
assert val1 is val2

E1-Bart0: Oh. It was in descriptions, sorry. But of course, it is good, because I came to it myself.

All_home_works/hw2/task4.py
Cjkjvfnby: Let's find another example of when this approach failes.

All_home_works/hw2/task4.py
```diff
@@ -0,0 +1,36 @@
...
+
+"""
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    memo = {}
+
+    def wrapper(*args, **kwargs):
+        call = (args, tuple(kwargs))
```
Cjkjvfnby, line 28: How about mutable arguments, why do you droped their support?

If we use only immutable arguments what could be wrong?  (give me 2 examples of different `*args`, `**kwargs` which will be different by will produce that same `call`) (Find it, add tests for it, fix it)

Cjkjvfnby (changes requested):

All_home_works/hw2/task4.py
E1-Bart0: I made mistake, i wanted: call = (args, tuple(kwargs.items())).
In the previous code kwargs not worked at all. And of course the problem with mutable arguments(

For this code: call = (args, tuple(kwargs.items())).
foo(1, 2.0, 0)
foo(True, 2, False)

And except for such easy examples, unfortunately, I can't suggest more examples.

Cjkjvfnby (approved):

PR #12, opened 2021-04-17
Hw3 task1

E1-Bart0:
Homework 3 task1 Done

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
     hooks:
       - id: black
   - repo: https://gitlab.com/pycqa/flake8/
-    rev: 3.9.1
+    rev: 3.9.0
     hooks:
       - id: flake8
         additional_dependencies:
@@ -40,6 +40,6 @@ repos:
     hooks:
       - id: isort
   - repo: https://github.com/jorisroovers/gitlint
-    rev: v0.15.1
+    rev: v0.15.0
     hooks:
       - id: gitlint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
     hooks:
       - id: black
   - repo: https://gitlab.com/pycqa/flake8/
-    rev: 3.9.0
+    rev: 3.9.1
     hooks:
       - id: flake8
         additional_dependencies:
@@ -40,6 +40,6 @@ repos:
     hooks:
       - id: isort
   - repo: https://github.com/jorisroovers/gitlint
-    rev: v0.15.0
+    rev: v0.15.1
     hooks:
       - id: gitlint
--- a/All_home_works/hw3/task1.py
+++ b/All_home_works/hw3/task1.py
@@ -26,22 +26,23 @@ f()
 
 def cache(times=0):
     def wrapper(func):
-        memo = {"result": None, "counter": times + 1}
+        memo = {}
 
         def inner_wrapper(*args, **kwargs):
-            _check_counter(memo, times)
+            call = frozenset(args) | frozenset(kwargs)
 
-            if memo["counter"] == times:
-                memo["result"] = func(*args, **kwargs)
+            if not _check_call_in_memo(call, memo):
+                memo[call] = {"result": func(*args, **kwargs), "counter": times}
 
-            return memo["result"]
+            return memo[call]["result"]
 
         return inner_wrapper
 
     return wrapper
 
 
-def _check_counter(memo, times):
-    if memo["counter"] < 1:
-        memo["counter"] = times + 1
-    memo["counter"] -= 1
+def _check_call_in_memo(call, memo):
+    if call in memo and memo[call]["counter"] > 0:
+        memo[call]["counter"] -= 1
+        return True
+    return False
--- a/All_home_works/hw3/task1.py
+++ b/All_home_works/hw3/task1.py
@@ -1,0 +1,48 @@
+"""
+In previous homework task 4, you wrote a cache function that remembers other function output value.
+ Modify it to be a parametrized decorator, so that the following code:
+
+ @cache(times=3)
+def some_function():
+    pass
+Would give out cached value up to times number only. Example:
+
+@cache(times=2)
+def f():
+    return input('? ')   # careful with input() in python2, use raw_input() instead
+
+f()
+? 1
+'1'
+f()     # will remember previous value
+'1'
+f()     # but use it up to two times only
+'1'
+f()
+? 2
+'2'
+"""
+
+
+def cache(times=0):
+    def wrapper(func):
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -20,7 +20,7 @@ repos:
     hooks:
       - id: black
   - repo: https://gitlab.com/pycqa/flake8/
-    rev: 3.9.0
+    rev: 3.9.1
```
Cjkjvfnby: It is not part of homework.  You don't need to show these changes to me, please remove them from PR.  

Technical things could be merged without involving me.

Cjkjvfnby (changes requested):
Let's get back to it after we will finish with the previous cache task.

Cjkjvfnby (changes requested):
Please don't assign me as a reviewer if the check failed.  I don't review such PR.

image

test/hw3/test_task1.py
```diff
@@ -0,0 +1,18 @@
...
+
+@cache(times=N)
+def f(attr=None):
+    return attr
+
+
+def test__cache():
+    input_data = 1
+    solution = f(input_data)
+    assert solution is f()
```
Cjkjvfnby: What are you testing in this line?

Cjkjvfnby (changes requested):

test/hw3/test_task1.py
```diff
@@ -0,0 +1,18 @@
...
+
+from All_home_works.hw3.task1 import cache
+
+
+@cache(times=2)
+def f():
+    return input("? ")
+
+
+@patch("hw3.test_task1.f", return_value=2)
```
Cjkjvfnby: Why do you need to patch function, when you can just decorate mock directly? 

mock = Mock()
function = cache(3)(mock)

test/hw3/test_task1.py
```diff
@@ -0,0 +1,18 @@
...
+def test__cache(f):
+    function = cache(3)(f)
+    function()
+    function()
+    function()
+    function()
+    assert 1 == len(f.mock_calls)
```
Cjkjvfnby: You intrigued me. You call function n times and stopped on a very interesting place.   What will happen next?

Cjkjvfnby: lets use: `asset_called_once_with`

Cjkjvfnby (changes requested):
You have added a single test for caching, lets also add a test that this function could work with different arguments.

test/hw3/test_task1.py
E1-Bart0: You intrigued me. You call function n times and stopped on a very interesting place. What will happen next?

Function "f" will call one more time. As counter in decorator became zero. And len(f.mock_call) will become 2

Cjkjvfnby: Please use quote when quoting 

image

Cjkjvfnby: > Function "f" will call one more time. As counter in decorator became zero. And len(f.mock_call) will become 2

Can you make a solid prove?

Cjkjvfnby (changes requested):
https://github.com/E1-Bart0/homework-repo/pull/12#pullrequestreview-645048328

test/hw3/test_task1.py
E1-Bart0: Снимок экрана 2021-04-29 в 14 38 13

E1-Bart0: Ok. I understood this task differently. Is it needed to remake?
An example from the description not very useful. It will be good if they will use your example, it's more informative.

Cjkjvfnby: > Ok. I understood this task differently. Is it needed to remake?

Yes. 

> An example from the description not very useful. 

This often happens in real life.  

The situation when you try, fail, think, try fail ....  and finally get the right unswear is very cool for getting experience.  
So having a bit vague task description is not a problem.

test/hw3/test_task1.py
```diff
@@ -0,0 +1,17 @@
+from unittest.mock import Mock, call
+
+from All_home_works.hw3.task1 import cache
+
+
+def test__cache_decorator():
```
Cjkjvfnby, line 6: Please add a test where you check the results of the function calls.  you can use `operator.add` as base functions.

Cjkjvfnby (changes requested):

test/hw3/test_task1.py
E1-Bart0: I do not understand clearly what do you want from me. And why it should be a different test, it's almost the same as the previous test, isn't it?

Cjkjvfnby: > it's almost the same as the previous test, isn't it?

Yes, it is.  I want you to test this "almost" part.  This small difference is one of the critical parts of the requirements: `Would give out cached value`

test/hw3/test_task1.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test__check_the_result_of_the_function_calls():
+    inner_function = Mock()
+    inner_function_with_decorator = cache(times=2)(inner_function)
+    function_calls = []
+
+    inner_function_with_decorator(1)
+    function_calls.append(call(1))
+    assert inner_function.call_args_list == function_calls
```
Cjkjvfnby: I'd prefer to remove `function_calls` and create it each time. 

```python
== []
== [call(1)]
== [call(1), call(2)]
```

It is really hard to track what you are expectation in this code.

Cjkjvfnby (changes requested):

E1-Bart0:
Maybe I do not clearly understand it too =)

> Please add a test where you check the results of the function calls. you can use operator.add as base functions.

All_home_works/hw3/task1.py
```diff
@@ -0,0 +1,48 @@
...
+    def wrapper(func):
+        memo = {}
+
+        def inner_wrapper(*args, **kwargs):
+            call = frozenset(args) | frozenset(kwargs)
+
+            if not _check_call_in_memo(call, memo):
+                memo[call] = {"result": func(*args, **kwargs), "counter": times}
+
+            return memo[call]["result"]
```
Cjkjvfnby, line 37: Let's try to remove this line.

Cjkjvfnby (changes requested):

All_home_works/hw3/task1.py
E1-Bart0: Are you trying to say that we do not check the function result in tests or need to reformat this code `memo[call]['result']` maybe with use walrus operator in if statement? Because we can't get rid of `return`

Cjkjvfnby: > Are you trying to say that we do not check the function result in tests 

Exactly.

Cjkjvfnby (approved):

PR #13, opened 2021-04-17
Hw3 task2

E1-Bart0:
Homework 3 task2 Done

### DIFF
--- a/All_home_works/hw3/task2.py
+++ b/All_home_works/hw3/task2.py
@@ -18,23 +18,11 @@ Calculate total sum of slow_calculate() of all numbers starting from 0 to 500.
   You are not allowed to modify slow_calculate function.
 """
 import hashlib
+import os
 import random
 import struct
 import time
 from multiprocessing import Pool
-
-import matplotlib.pyplot as plt
-
-
-def timer(func):
-    def wrapper(num):
-        start = time.time()
-        res = func(num)
-        finish = time.time() - start
-        finish_per_number = finish / num
-        return res, finish, finish_per_number
-
-    return wrapper
 
 
 def slow_calculate(value):
@@ -43,25 +31,7 @@ def slow_calculate(value):
     return sum(struct.unpack("<" + "B" * len(data), data))
 
 
-@timer
 def sum_of_slow_calculate(max_value=500):
-    pool = Pool(processes=max_value)
-    array = pool.map(slow_calculate, range(max_value))
+    pool = Pool(processes=os.cpu_count())
+    array = pool.map(slow_calculate, range(max_value + 1))
     return sum(array)
-
-
-def get_data(data):
-    res = []
-    [res.append(sum_of_slow_calculate(i)[1:]) for i in data]
-    return res
-
-
-def plot():
-    x = list(range(1, 50)) + list(range(50, 500, 50))
-    y = get_data(x)
-    fig, ax = plt.subplots(figsize=(8, 4))
-    ax.plot(
-        x,
-        y,
-    )
-    plt.show()
--- a/All_home_works/hw3/task2.py
+++ b/All_home_works/hw3/task2.py
@@ -25,8 +25,8 @@ from multiprocessing import Pool
 
 
 def slow_calculate(value):
-    time.sleep(random.randint(1, 3))
-    data = hashlib.md5(str(value).encode()).digest()
+    time.sleep(random.randint(1, 3))  # noqa: S311
+    data = hashlib.md5(str(value).encode()).digest()  # noqa: S303
     return sum(struct.unpack("<" + "B" * len(data), data))
 
 
--- a/All_home_works/hw3/task2.py
+++ b/All_home_works/hw3/task2.py
@@ -1,0 +1,47 @@
+"""
+Here's a not very efficient calculation function that calculates something important:
+
+import time
+import struct
+import random
+import hashlib
+
+def slow_calculate(value):
+
+    time.sleep(random.randint(1,3))
+    data = hashlib.md5(str(value).encode()).digest()
+    return sum(struct.unpack('<' + 'B' * len(data), data))
+
+Calculate total sum of slow_calculate() of all numbers starting from 0 to 500.
+ Calculation time should not take more than a minute.
+ Use functional capabilities of multiprocessing module.
+  You are not allowed to modify slow_calculate function.
+"""
+import hashlib
+import random
+import struct
+import time
+from multiprocessing import Pool
+
+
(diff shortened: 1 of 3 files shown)


test/hw3/test_task2.py
```diff
@@ -0,0 +1,14 @@
+from All_home_works.hw3.task2 import sum_of_slow_calculate
+
+
+def test_sum_of_slow_calculate():
+    expected = 1024259
+    max_time = 60
```
Cjkjvfnby: You don't need to test time with unittests.  Just do it on your machine.

All_home_works/hw3/task2.py
```diff
@@ -0,0 +1,47 @@
...
+        result = func(*args, **kwargs)
+        delta = time.time() - start
+        return result, delta
+
+    return wrapper
+
+
+@timer
+def sum_of_slow_calculate(max_value=500):
+    pool = Pool(processes=max_value)
```
Cjkjvfnby, line 34: Is it a recommended way to use Pool?

Cjkjvfnby (changes requested):

All_home_works/hw3/task2.py
E1-Bart0: Nope. I forgot about the context manager.
By the way, I have a question. Maybe better to use threading if we have 500 processes?

Cjkjvfnby: Threading will not work, you will get an explanation of why in the corresponding lecture.

Why do we need to have 500 processes?

All_home_works/hw3/task2.py
```diff
@@ -0,0 +1,36 @@
...
+import random
+import hashlib
+
+def slow_calculate(value):
+
+    time.sleep(random.randint(1,3))
+    data = hashlib.md5(str(value).encode()).digest()
+    return sum(struct.unpack('<' + 'B' * len(data), data))
+
+Calculate total sum of slow_calculate() of all numbers starting from 0 to 500.
```
Cjkjvfnby, line 15: How many numbers do you need to calculate?

All_home_works/hw3/task2.py
E1-Bart0: It's actually not needed. But, I thought, in multiprocessing, it's recommended to use such many processes as we have cores in our system.

Cjkjvfnby: >  it's recommended to use such many processes as we have cores in our system.

It's ok to have 500 cores on your machine. But GitHub actions are run on less powerful machines, let's set more reasonable numbers. 

As well please do some experiments and check the number of processes that will give the best performance on your machine (binary search, 1 execution per number, we don't need high accuracy)

Cjkjvfnby (changes requested):

All_home_works/hw3/task2.py
E1-Bart0: So, I have 4 core. That's why time at first not increased. As all cores worked simultaneously. Then time began to slow down with increasing processes because the core always had to do many switches from one activity to another.
image

Cjkjvfnby: How can we set a proper number of cores for the execution on different machines?

Cjkjvfnby: You have answered another comment.  This makes our conversation a bit messy.  Don't forget to answer to initial comment too.

E1-Bart0: How many numbers do you need to calculate? 
From 0 to 500 -> 501 number or what? Sorry, I do not understand you.

How can we set a proper number of cores for the execution on different machines?
os.cpu_count() return amount of cores in our system

Cjkjvfnby: >  From 0 to 500 -> 501 number or what? Sorry, I do not understand you.

This is a leading question. In that case, I want to bring your attention to the usage of that number. 

> os.cpu_count() return amount of cores in our system

Is it possible to use it?  (This is another leading questions, in this time it is "call to action")

E1-Bart0: Totally did not understand you. Not needed to use 500 or 501 calculations. Needed to use as many cores as cores in the system and to checking cores by os.cpu_count(). Am I right?

Cjkjvfnby: > Not needed to use 500 or 501 calculations.

What is written in the requirements about that?

> Needed to use as many cores as cores in the system and to checking cores by os.cpu_count(). Am I right?

Yes.

Cjkjvfnby (changes requested):

All_home_works/hw3/task2.py
E1-Bart0: But the task is: "Calculate the total sum of slow_calculate() **of all numbers** starting from 0 to 500."

And one more line: "Calculation time should not take more than a minute."

Even without multiprocessing, I will get a result from 0 to os.cpu_count() for this time...

E1-Bart0: As always, I do not understand. That you speaking about Pool. My system works fine with 500 pools, but I will make as you said.

E1-Bart0:
On one system with mac and the other with Linux, it works much faster with Pool(500) than Pool(os.cpu_count()).

Cjkjvfnby:
> On one system with mac and the other with Linux, it works much faster with Pool(500) than Pool(os.cpu_count()).

Is it the same as https://github.com/E1-Bart0/homework-repo/pull/13#discussion_r619840424?

All_home_works/hw3/task2.py
Cjkjvfnby: > Even without multiprocessing, I will get a result from 0 to os.cpu_count() for this time...

The number of calculations and the number of cores are not related.  They are different arguments and changed for different reasons.  In some cases it might look like they are related.

E1-Bart0:
> > On one system with mac and the other with Linux, it works much faster with Pool(500) than Pool(os.cpu_count()).
> 
> Is it the same as [#13 (comment)](https://github.com/E1-Bart0/homework-repo/pull/13#discussion_r619840424)?

I cannot check it now, because of weekends. But with os.cpu_count() it had been calculating rather longer than a minute.

Cjkjvfnby:
> I cannot check it now, because of weekends. But with os.cpu_count() it had been calculating rather longer than a minute.

This is a sample of a hardware-related problem. Different OS and processor gives a different result.  This code works faster on your machine and fails on mine.  It is a very complex task to make it work fast everywhere.

The task accepted.

Cjkjvfnby (approved):

PR #14, opened 2021-04-17
Hw3 task3

E1-Bart0:
Homework 3 task3 Done

### DIFF
--- a/All_home_works/hw3/task3.py
+++ b/All_home_works/hw3/task3.py
@@ -11,7 +11,7 @@ class Filter:
         self.functions = functions
 
     def apply(self, data):
-        return [item for item in data if all(i(item) for i in self.functions[0])]
+        return [item for item in data if all(i(item) for i in self.functions)]
 
 
 # example of usage:
@@ -30,7 +30,7 @@ def make_filter(**kwargs):
             return checking_dict[key] == value if key in checking_dict else False
 
         filter_funcs.append(keyword_filter_func)
-    return Filter(filter_funcs)
+    return Filter(*filter_funcs)
 
 
 sample_data = [
--- a/All_home_works/hw3/task3.py
+++ b/All_home_works/hw3/task3.py
@@ -33,16 +33,6 @@ def make_filter(**kwargs):
     return Filter(*filter_funcs)
 
 
-sample_data = [
-    {
-        "name": "Bill",
-        "last_name": "Gilbert",
-        "occupation": "was here",
-        "type": "person",
-    },
-    {"is_dead": True, "kind": "parrot", "type": "bird", "name": "polly"},
-]
-
 # make_filter(name='polly', type='bird').apply(sample_data) should return only second entry from the list
 
 # There are multiple bugs in this code. Find them all and write tests for faulty cases.
--- a/All_home_works/hw3/task3.py
+++ b/All_home_works/hw3/task3.py
@@ -7,27 +7,27 @@ class Filter:
     functions that return True if object in list conforms to some criteria
     """
 
-    def __init__(self, functions):
+    def __init__(self, *functions):
         self.functions = functions
 
     def apply(self, data):
-        return [item for item in data if all(i(item) for i in self.functions)]
+        return [item for item in data if all(i(item) for i in self.functions[0])]
 
 
 # example of usage:
-# positive_even = Filter(lamba a: a % 1 == 0, lambda a: a > 0, lambda a: isinstance(int, a)))
+# positive_even = Filter(lambda a: a % 1 == 0, lambda a: a > 0, lambda a: isinstance(int, a)))
 # positive_even.apply(range(99)) should return only even numbers from 0 to 99
 
 
-def make_filter(**keywords):
+def make_filter(**kwargs):
     """
     Generate filter object for specified keywords
     """
     filter_funcs = []
-    for key, value in keywords.items():
+    for key, value in kwargs.items():
 
-        def keyword_filter_func(value):
-            return value[key] == value
+        def keyword_filter_func(checking_dict):
+            return checking_dict[key] == value if key in checking_dict else False
 
         filter_funcs.append(keyword_filter_func)
     return Filter(filter_funcs)
--- a/All_home_works/hw3/task3.py
+++ b/All_home_works/hw3/task3.py
@@ -1,0 +1,48 @@
+# I decided to write a code that generates data filtering object from a list of keyword parameters:
+
+
+class Filter:
+    """
+    Helper filter class. Accepts a list of single-argument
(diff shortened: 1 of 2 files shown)


Cjkjvfnby (changes requested):
There are examples of code, why not use them?

E1-Bart0:
Is it needed to rewrite?

Cjkjvfnby:
> Is it needed to rewrite?

Why are you asking questions instead of action? Have you tried to use examples? Did they succeed?

E1-Bart0:
I haven't used such examples, because there are a few mistakes in them:
positive_even = Filter(lambda a: a % 1 == 0, lambda a: a > 0, lambda a: isinstance(int, a)))
lambda a: a % 1 == 0. It's always True, needed a % 2 == 0 for even numbers
lambda a: isinstance(int, a). Needed isinstance(a, int)

So I asked you why it's needed to rewrite? Because I have some misunderstanding from previous dialogs.

test/hw3/test_task3.py
```diff
@@ -0,0 +1,36 @@
+from All_home_works.hw3.task3 import Filter, make_filter, sample_data
+
+
+def test_filter():
+    expected = list(range(2, 100, 2))
```
Cjkjvfnby: Why do you need to have 50 elements for tests?

test/hw3/test_task3.py
```diff
@@ -0,0 +1,36 @@
...
+    positive_even = Filter(
+        lambda a: a % 2 == 0, lambda a: a > 0, lambda a: isinstance(a, int)
+    )
+    result = positive_even.apply(range(99))
+    assert expected == result
+
+
+def test_filter_case__data_with_negative():
```
Cjkjvfnby: What is negative?  This name does not describe the purpose of this test well.

test/hw3/test_task3.py
```diff
@@ -0,0 +1,36 @@
...
+    expected = list(range(2, 100, 2))
+    positive_even = Filter(
+        lambda a: a % 2 == 0, lambda a: a > 0, lambda a: isinstance(a, int)
+    )
+    result = positive_even.apply(range(-99, 99))
+    assert expected == result
+
+
+def test_filter_make_functions_valid_for_second():
```
Cjkjvfnby: Why for a second? Why not for a minute?

All_home_works/hw3/task3.py
```diff
@@ -0,0 +1,48 @@
...
+    for key, value in kwargs.items():
+
+        def keyword_filter_func(checking_dict):
+            return checking_dict[key] == value if key in checking_dict else False
+
+        filter_funcs.append(keyword_filter_func)
+    return Filter(*filter_funcs)
+
+
+sample_data = [
```
Cjkjvfnby: This is example data, no need to leave it here with function. Just move it to tests.

test/hw3/test_task3.py
```diff
@@ -0,0 +1,36 @@
...
+    expected = sample_data[1]
+    result = make_filter(name="polly", type="bird").apply(sample_data)
+    assert expected == result[0]
+
+
+def test_filter_make_functions_valid_for_first():
```
Cjkjvfnby: If we filter the dictionary by key-value which negative cases we could tests?

Cjkjvfnby (changes requested):
> So I asked you why it's needed to rewrite? Because I have some misunderstanding from previous dialogs.

It was about supporting multiple filters at once.

Cjkjvfnby (changes requested):
Please address all comments.

test/hw3/test_task3.py
E1-Bart0: Fixed

All_home_works/hw3/task3.py
E1-Bart0: Fixed

test/hw3/test_task3.py
E1-Bart0: Rename: test_filter__positive_even__data_with_negative_numbers_in_data

test/hw3/test_task3.py
E1-Bart0: Rename: test_filter__make_functions__valid_for_only_one_dict

test/hw3/test_task3.py
E1-Bart0: test_filter__make_functions__valid_for_only_one_dict,
test_filter__make_functions__valid_for_all_dict,
test_filter__make_functions__not_valid_for_all_dict,
test_filter__make_functions__not_valid_for_all_dict_with_not_existing_key

All_home_works/hw3/task3.py
Cjkjvfnby: Have you moved it?

test/hw3/test_task3.py
```diff
@@ -0,0 +1,53 @@
...
+    data = [-2 - 1, 0, 1, 2, 3, 4]
+    expected = [2, 4]
+    positive_even = Filter(
+        lambda a: a % 2 == 0, lambda a: a > 0, lambda a: isinstance(a, int)
+    )
+    result = positive_even.apply(data)
+    assert expected == result
+
+
+def test_filter__make_functions__valid_for_only_one_dict():
```
Cjkjvfnby: What is valid? This term is not used in the context of filtering.

> `for_only_one_dict` 

There is always only one dict in your tests. Probably you mean item.

Cjkjvfnby (changes requested):

All_home_works/hw3/task3.py
E1-Bart0: I copied not moved it into the test because it was in a description and I did not want to touch it...
Now this data only in the test.

Cjkjvfnby: > I copied not moved it into the test because it was in a description and I did not want to touch it...

If I ask you something and you disagree, it is ok to say that. When you respond "fixed" and do something else, it looks very strange.

test/hw3/test_task3.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_filter__make_functions__not_valid_for_items():
+    result = make_filter(valid=False).apply(TEST_DATA)
+    assert [] == result
+
+
+def test_filter__make_functions__not_valid_for_all_item_with_not_existing_key():
+    result = make_filter(not_valid=True).apply(TEST_DATA)
+    assert not result
```
Cjkjvfnby: In another test you use.
```python
assert [] == result
```

It is better to use consistent approach for same things.

Cjkjvfnby (approved):

test/hw3/test_task3.py
E1-Bart0: Will remember!

PR #15, opened 2021-04-17
"hw3 task4"

E1-Bart0:
Homework 3 task4 Done

### DIFF
--- a/test/hw3/test_task4.py
+++ b/test/hw3/test_task4.py
@@ -1,0 +1,19 @@
+from All_home_works.hw3.task4 import is_armstrong
+
+
+def test_is_armstrong_case1():
+    assert is_armstrong(153)
+
+
+def test_is_armstrong_case2():
+    assert not is_armstrong(10)
+
+
+def test_is_armstrong_case3():
+    assert is_armstrong(9)
+
+
+if __name__ == "__main__":
+    test_is_armstrong_case1()
+    test_is_armstrong_case2()
+    test_is_armstrong_case3()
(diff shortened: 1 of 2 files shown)


test/hw3/test_task4.py
```diff
@@ -0,0 +1,19 @@
...
+
+def test_is_armstrong_case2():
+    assert not is_armstrong(10)
+
+
+def test_is_armstrong_case3():
+    assert is_armstrong(9)
+
+
+if __name__ == "__main__":
```
Cjkjvfnby: Why did you need an additional test runner?

test/hw3/test_task4.py
```diff
@@ -0,0 +1,19 @@
+from All_home_works.hw3.task4 import is_armstrong
+
+
+def test_is_armstrong_case1():
```
Cjkjvfnby: `case1` does not mean anything to me.  Please give a more meaningful name for the test.

Cjkjvfnby (changes requested):

test/hw3/test_task4.py
E1-Bart0: This case is in an Example: "
Examples:
- 9 is an Armstrong number, 9 = 9^1 = 9
- 10 is not: 10 != 1^2 + 0^2 = 1
- 153 is : 153 = 1^3 + 5^3 + 3^3 = 1 + 125 + 27 = 153
".
So I decided to check it.

Cjkjvfnby: `if __name__ == "__main__":`  is for executable files, you have have test runner you don't need it.  

Please make sure that there is not such line in the tests of the other tasks.

Cjkjvfnby (approved):

PR #18, opened 2021-04-28
Hw4 task2

E1-Bart0:


All_home_works/hw4/task2.py
```diff
@@ -0,0 +1,30 @@
...
+* https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen
+"""
+from urllib import request
+
+
+def count_dots_on_i(url: str) -> int:
+    with request.urlopen(url) as html:  # noqa: S310
+        counter = 0
+        for line in html:
+            counter += sum(1 for char in line.decode("utf-8") if char == "i")
```
Cjkjvfnby, line 29: Can we simplify it?

test/hw4/test_task2.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def __enter__(self):
+        return map(lambda x: x.encode(), self.data)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        return None
+
+
+@patch("urllib.request.urlopen")
+def test_urlopen__with_right_url(urlopen):
```
Cjkjvfnby, line 20: You can join it with `test_count_dots_on_i__check_valid_data`

- `@patch(` is the preparation step (`Given`)
- `count_dots_on_i` is our action (`When`)
- mock call and return value are the result of the action that should be asserted (`Then`). 

In that case your action is complex and this is ok to test all it's aspects in the single test.

test/hw4/test_task2.py
```diff
@@ -0,0 +1,38 @@
+from unittest.mock import patch
+
+import pytest
+
+from All_home_works.hw4.task2 import count_dots_on_i
+
+URL = "https://example.com/"
```
Cjkjvfnby: Please inline it, and put something that does not look like real URL, to be sure that it will not be called.

Cjkjvfnby (approved):

PR #19, opened 2021-04-28
Hw4 task3

E1-Bart0:


test/hw4/test_task3.py
```diff
@@ -0,0 +1,28 @@
+from All_home_works.hw4.task3 import my_precious_logger
+
+ERROR = "error: 404"
+TEXT = "OK: 200"
```
Cjkjvfnby: It is better to inline these variables.

test/hw4/test_task3.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def test_my_precious_logger__error_msg__test_msg_in_stderr(capsys):
```
Cjkjvfnby: It's ok to join this test with test_my_precious_logger__error_msg__test_none_in_stdout

Cjkjvfnby (changes requested):
Let's try to find a bit more cases we could test here.

Cjkjvfnby (changes requested):
https://github.com/E1-Bart0/homework-repo/pull/19#pullrequestreview-649813423

test/hw4/test_task3.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_my_precious_logger__ok__msg__test_msg_in_stdout(capsys):
+    text = "OK: 200"
+    my_precious_logger(text=text)
+    captured = capsys.readouterr()
+    assert captured.out == text
+    assert not captured.err
+
+
+def test_my_precious_logger__ok__msg_start_as_ordinary_msg__test_msg_in_stdout(capsys):
```
Cjkjvfnby: What is the difference between Ok and ordinary cases?

Cjkjvfnby (changes requested):
Let's add one more positive test.   hint: the key is in the task requirements.

test/hw4/test_task3.py
E1-Bart0: I think it's not actually needed to test.

test/hw4/test_task3.py
```diff
@@ -0,0 +1,29 @@
...
+
+def test_my_precious_logger__error_msg__test_msg_in_stderr(capsys):
+    error = "error: 404"
+    my_precious_logger(text=error)
+    captured = capsys.readouterr()
+    assert captured.err == error
+    assert not captured.out
+
+
+@pytest.mark.parametrize("input_data", ["Error: 404", "error 404"])
```
Cjkjvfnby: Almost there, need to find one more combination with `error`.  There is a keyword for it in the requirements. 

Also, this test helped me to find a bug in your code.  One of the `input_data` should be in stderr.

Cjkjvfnby (changes requested):

test/hw4/test_task3.py
E1-Bart0: Task description:

> if line starts with "error"

My mistake is: 'error:'

And next combination is: 'errorTEXT' or 'errorerror'?

Cjkjvfnby: > And next combination is: 'errorTEXT' or 'errorerror'?

No. 

This is in that phrase `if line starts with "error"`  and it is not `error`.

E1-Bart0: We are not considering bytes string? I do not know.

Cjkjvfnby: > We are not considering bytes string? 

This looks like a randomly dropped suggestion.  I don't see any hints in the phrase `if a line starts with "error"` which could lead to bytes string.   

There is a technique named brainstorming when you throw some raw ideas. But brainstorming is not used in the current context. In teamwork, the fast and random answer gives more problems than it solves. Just simple `I do not know` would be much better. 

You need to find some cases that could violate this requirement:
```
if line starts with "error"
```

E1-Bart0: I'm sorry, but I do not know...

Cjkjvfnby: You have tested `if`, you have tested `error` what else has left?

E1-Bart0: if "error" in line, but not line startswith "error"

Cjkjvfnby: Exactly!

Cjkjvfnby (approved):

PR #22, opened 2021-04-29
"Hw5 task2"

E1-Bart0:


test/hw5/test_task2.py
```diff
@@ -0,0 +1,15 @@
+from All_home_works.hw5.task2 import print_result
+
+
+def foo():
+    """DOCS OF FOO"""
```
Cjkjvfnby: Please follow docstring convention.

test/hw5/test_task2.py
```diff
@@ -0,0 +1,15 @@
...
+    return "return foo"
+
+
+def test_doc_string():
+
+    foo_with_decorator = print_result(foo)
+    foo_with_decorator()
```
Cjkjvfnby: Do you need to call it?

Cjkjvfnby (changes requested):

test/hw5/test_task2.py
E1-Bart0: I made mistake. I appropriated doc and name before the function had been ever called which explains why it worked only after called.

Cjkjvfnby: > which explains why it worked only after called

It's worth exploring such strange cases before requesting a review.

Cjkjvfnby (changes requested):
I think we could add one more test.

Cjkjvfnby (approved):

PR #23, opened 2021-05-04
Hw6 task1

E1-Bart0:


Cjkjvfnby (changes requested):
You have implemented the task requirements. 

But let's try to find some corner cases when this does not work.

E1-Bart0:
Are you talking about instances_counter or about class?
Are the cases when this does not work connected with class __new__ method?

Cjkjvfnby:
>  Are the cases when this does not work connected with class new method?

Kind of.     What else we could do with a class besides instancination?

test/hw6/test_task1.py
```diff
@@ -0,0 +1,107 @@
...
+    return TestClass
+
+
+def test__get_created_instances__check_from_class(test_class):
+    assert 0 == test_class.get_created_instances()
+    test_class()
+    assert 1 == test_class.get_created_instances()
+
+
+def test__get_created_instances__check_method_from_class_instance(test_class):  # noqa:
```
Cjkjvfnby: `# noqa:` ?

test/hw6/test_task1.py
```diff
@@ -0,0 +1,107 @@
...
+    class A:
+        @staticmethod
+        def get_created_instances():
+            return "foo"
+
+        @staticmethod
+        def reset_instances_counter():
+            return "foofoo"
+
+    _, _, _ = A(), A(), A()
```
Cjkjvfnby: Why do you need this line?

test/hw6/test_task1.py
```diff
@@ -0,0 +1,107 @@
...
+    class A(test_class):
+        pass
+
+    _, _, _ = A(), A(), A()
+    assert 3 == A.get_created_instances()
+    A.reset_instances_counter()
+    assert 0 == A.get_created_instances()
+
+
+def test__instances_counter_breaks_class_methods():
```
Cjkjvfnby: `breaks` -> `overwrites`  methods are not broken, they just replaced.

test/hw6/test_task1.py
```diff
@@ -0,0 +1,107 @@
...
+        def reset_instances_counter():
+            return "foofoo"
+
+    _, _, _ = A(), A(), A()
+    assert "foo" == A.get_created_instances()
+    assert "foofoo" == A.reset_instances_counter()
+
+
+def test__instances_counter_breaks_metaclass_if_metaclass_is_defining_method_for_class_which_are_decorators_method():
+    class B(type):
```
Cjkjvfnby: Actually, this is the same test as for an inheritance. 

Metaclass construct class before it is passed to our decorator.

test/hw6/test_task1.py
```diff
@@ -0,0 +1,107 @@
...
+def test__instances_counter__check_class_instance_not_broken(test_class):
+    test = test_class(var=10)
+    assert 1 == test.attribute
+    test.add_one()
+    assert 2 == test.attribute
+    assert 10 == test.var
+
+
+def test__instances_counter__check_inherit_from_class_with_decorator(test_class):
+    class A(test_class):
```
Cjkjvfnby, line 49: Let's play with the initializer.  The different number of arguments, super and not super.  

Maybe some tests will require `pytest.raises`. We are documenting the behavior of the decorator.

Cjkjvfnby (changes requested):

test/hw6/test_task1.py
```diff
@@ -81,27 +87,26 @@ def get_created_instances():
...
+            self.reset_instances_counter = reset_instances_counter
+            self.get_created_instances = get_created_instances
+
+    test_class = Test("reset", "get")
+    assert test_class.get_created_instances == "get"
+    assert test_class.reset_instances_counter == "reset"
+
+    assert 1 == Test.get_created_instances()
+    with pytest.raises(TypeError, match="'str' object is not callable"):
+        test_class.get_created_instances()
```
Cjkjvfnby: You can check that attribute is a string with less code.

Cjkjvfnby (approved):

PR #25, opened 2021-05-05
Hw7 task1

E1-Bart0:

PR #28, opened 2021-05-05
Hw7 task2

E1-Bart0:


test/hw7/test_task2.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from All_home_works.hw7.task2 import backspace_compare
+
+
+@pytest.mark.parametrize(  # noqa: PT006
```
Cjkjvfnby: What is PT006?

test/hw7/test_task2.py
```diff
@@ -0,0 +1,26 @@
...
+    "first,second,expected",
+    [
+        ("aB#c", "aD#c", True),
+        ("A##c", "#A#c", True),
+        ("#", "", True),
+        ("abcABCD####", "abc", True),
+    ],
+)
+def test__backspace_compare__with_backspace__expected_true(first, second, expected):
```
Cjkjvfnby: You have `expected_true` in your name. That means that you could simplify this test a bit.

Cjkjvfnby (changes requested):
The solution is good, need to update tests a bit.

test/hw7/test_task2.py
E1-Bart0: https://github.com/E1-Bart0/homework-repo/pull/25#discussion_r627397514

test/hw7/test_task2.py
E1-Bart0: I have a question to ask you.
If we have True it's recommended to just assert, but if False or None it's better to use assert False and assert None, isn't it?

Cjkjvfnby: > If we have True it's recommended to just assert

Sometimes it is ok to do `assert result is True` just to clarify that you need boolean True explicitly.  But it is not critical. 

After that you have changed the assertion statement, what could you remove?

Cjkjvfnby (approved):
Bonus task: write this using generator. (O(1) memory)

All_home_works/hw7/task2.py
```diff
@@ -13,18 +13,24 @@
...
+from itertools import zip_longest
 
 
 def backspace_compare(first: str, second: str) -> bool:
-    return get_filtered(first) == get_filtered(second)
+    first_gen = get_filtered(first)
+    second_gen = get_filtered(second)
+    return all(
+        char1 == char2
+        for char1, char2 in zip_longest(first_gen, second_gen, fillvalue=False)
```
Cjkjvfnby: `fillvalue=False` it's better to use the value of the same type.

Cjkjvfnby (approved):

PR #30, opened 2021-05-06
Hw8_task1

E1-Bart0:


All_home_works/hw8/task1.py
```diff
@@ -0,0 +1,70 @@
...
+In case of attribute clash existing built-in attributes take precedence.
+In case when value cannot be assigned to an attribute
+(for example when there's a line 1=something) ValueError should be raised.
+File size is expected to be small, you are permitted to read it entirely into memory.
+"""
+import string
+
+
+class KeyValueStorage:
+    allowed_chars = set(string.digits) | set(string.ascii_letters) | set("_")
```
Cjkjvfnby: Let's use set literal here.

All_home_works/hw8/task1.py
```diff
@@ -0,0 +1,70 @@
...
+    def _get_data_from_file(self):
+        data = {}
+        for line in self._get_lines_from_file():
+            attr_name, attr_value = line.split("=")
+            attr_value = int(attr_value) if attr_value.isdigit() else attr_value
+            self._check_attr_name_on_syntax(attr_name)
+            data[attr_name] = attr_value
+        return data
+
+    def _check_attr_name_on_syntax(self, name):
```
Cjkjvfnby: Please check the batteries.

test/hw8/test_task1.py
```diff
@@ -0,0 +1,95 @@
...
+    assert storage["new_1"] == "NEW_1"
+    assert storage["int_value"] == 10
+
+
+def test_key_value_storage__not_existing_attributes(file_name):
+    storage = KeyValueStorage(file_name)
+    with pytest.raises(AttributeError, match="object has no attribute"):
+        error = storage.not_existing
+    with pytest.raises(AttributeError, match="object has no attribute"):
+        error = storage["not_existing"]  # noqa: F841
```
Cjkjvfnby: What is the recommended exception in that case?

test/hw8/test_task1.py
```diff
@@ -0,0 +1,95 @@
...
+):
+    with pytest.raises(ValueError, match="Syntax error: Not valid method name:"):
+        KeyValueStorage(file_name)
+
+
+@patch(
+    "All_home_works.hw8.task1.KeyValueStorage._get_lines_from_file",
+    return_value=("path=ERROR",),
+)
+def test_key_value_storage__if_attribute_already_exists(  # noqa: PT019
```
Cjkjvfnby: What is PT019?

test/hw8/test_task1.py
```diff
@@ -0,0 +1,95 @@
...
+    storage = KeyValueStorage(file_name)
+    assert storage.path == file_name
+    assert storage["path"] == "ERROR"
+
+
+@patch(
+    "All_home_works.hw8.task1.KeyValueStorage._get_lines_from_file",
+    return_value=("foo=1", "foo_method=1"),
+)
+def test_key_value_storage__if_method_already_exists(  # noqa: PT019
```
Cjkjvfnby: You could use exact phrases from the task description for tests.

Cjkjvfnby (changes requested):

All_home_works/hw8/task1.py
E1-Bart0: Are you talking about naming, because I don't understand batteries=)?

test/hw8/test_task1.py
E1-Bart0: ValueError

test/hw8/test_task1.py
E1-Bart0: I cut it off

All_home_works/hw8/task1.py
Cjkjvfnby: This is a reference to "batteries included".

test/hw8/test_task1.py
Cjkjvfnby: Could a provide a link to the explanation?

E1-Bart0: > In case when value cannot be assigned to an attribute
(for example when there's a line 1=something) ValueError should be raised.

So let cut brackets.
> In case when value cannot be assigned to an attribute ValueError should be raised.

And in my case, I couldn't assign value, because I do not have such an attribute. So in some sense, it should be a ValueError exception.

Cjkjvfnby: I am not sure what is `value` and what is `attribute` here.

E1-Bart0: I'm a little bit confusing. But if read the description with brackets for me it's better to use AttributeError, because we do not have such an attribute. But in the same way, it can be a ValueError because of description. So what do you think?

All_home_works/hw8/task1.py
E1-Bart0: Thanks! I Will remember it `str.isidentifier`.

test/hw8/test_task1.py
Cjkjvfnby: > description with brackets 

That thing has a name. 

> for me it's better to use AttributeError, because we do not have such an attribute

Please dig a bit into documentation and describe which syntax in Python leads to AttributeError.

> it can be a ValueError because of description.

Do you mean that because we should raise ValueError when we construct our class we also should raise it when we use our class?

Cjkjvfnby (changes requested):
Please fix items we have discassed on a call.

All_home_works/hw8/task1.py
```diff
@@ -50,15 +50,17 @@ def _get_data_from_file(self):
         return data
 
     def __getitem__(self, item):
-        return self._get_value_if_exists(item)
+        value = self.data.get(item, None)
+        if value is None:
+            raise KeyError(
```
Cjkjvfnby, line 55: You don't need to raise it yourself, you can just return self.data[item]

Cjkjvfnby (approved):

All_home_works/hw8/task1.py
E1-Bart0: But I wanted to have a more beautiful description of this error. Than:
 `value = self.data[item]`
` KeyError: 'not_existing'`

PR #31, opened 2021-05-06
Hw8_task2

E1-Bart0:


All_home_works/hw8/task2.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @staticmethod
+    def get_row_as_dict_from(cursor):
+        data = cursor.fetchone()
+        if data is None:
+            return None
+        column_name = (x[0] for x in cursor.description)
+        return {key: value for key, value in zip(column_name, data)}
+
+    def __del__(self):
```
Cjkjvfnby: When this method is called?

All_home_works/hw8/task2.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __len__(self):
+        self.cursor.execute(
+            f"SELECT count(*) FROM '{self.db_table}' query"  # noqa: S608
+        )
+        return self.cursor.fetchone()[0]
+
+    def __getitem__(self, item):
+        self.cursor.execute(
+            f"SELECT * FROM '{self.db_table}' where name='{item}'"  # noqa: S608
```
Cjkjvfnby: What could go wrong here?

All_home_works/hw8/task2.py
```diff
@@ -0,0 +1,64 @@
...
+
+class TableData:
+    def __init__(self, db_name, db_table):
+        self.conn = sqlite3.connect(db_name)
+        self.cursor = self.conn.cursor()
+        self.db_table = db_table
+
+    def __iter__(self):
+        self.cursor.execute(f"SELECT * FROM '{self.db_table}'")  # noqa: S608
+        return Data(cursor=self.cursor)  # noqa: S608
```
Cjkjvfnby: What is S608?

All_home_works/hw8/task2.py
```diff
@@ -0,0 +1,64 @@
...
+        self.cursor.execute(
+            f"SELECT count(*) FROM '{self.db_table}' query"  # noqa: S608
+        )
+        return self.cursor.fetchone()[0]
+
+    def __getitem__(self, item):
+        self.cursor.execute(
+            f"SELECT * FROM '{self.db_table}' where name='{item}'"  # noqa: S608
+        )
+        return self.get_row_as_dict_from(self.cursor)
```
Cjkjvfnby: What is another option to do it?

test/hw8/test_task2.py
```diff
@@ -0,0 +1,114 @@
...
+def test_table_data__get_row_from_db_by_name__check_storage_after_update(updated_db):
+    storage = updated_db
+    assert {"name": "Test Book", "author": "Test Author"} == storage["Test Book"]
+
+
+def test_table_data__for_loop(create_db_table):
+    db_name, db_table = create_db_table
+    storage = TableData(db_name, db_table)
+    iter_storage = iter(storage)
+    row = next(iter_storage)
```
Cjkjvfnby:
```
>>> x =  iter(storage)
>>> x
x
>>> list(x) 
[x, x]
```

I found it extremely confusing, that iterator returns instances of itself during iteration.
I expect that the iterator returns a new value each time.  Instead of that, it returns the same value and mutates it.

Please avoid any mutations after object was intialized.

Cjkjvfnby (changes requested):

All_home_works/hw8/task2.py
E1-Bart0: it's garbage collection. Are you trying to say use __open__, __exit__?

All_home_works/hw8/task2.py
E1-Bart0: S608 Possible SQL injection vector through string-based query construction

All_home_works/hw8/task2.py
E1-Bart0: - Maybe there is no such table
- Maybe there is no such row in the table
- Maybe there are some rows with such name, not one

test/hw8/test_task2.py
E1-Bart0: Because of the task. Needed that worked:
`'NAME' in storage`
And it will check __iter__ from the class. And __iter__ returns self, where __eq__ is overwrite just to find it.
if __iter__ will return a dict or a list: `'NAME' in storage` will return False, needed:
`if 'NAME' in (v[0] for v in storage)`

All_home_works/hw8/task2.py
Cjkjvfnby: > it's garbage collection

It is not an answer to the question I have asked. What is the correct answer?

> Are you trying to say use open, exit?

I am saying that resources should be properly closed after use.

All_home_works/hw8/task2.py
Cjkjvfnby: What else?

All_home_works/hw8/task2.py
Cjkjvfnby: Can we fix it in this line?

test/hw8/test_task2.py
Cjkjvfnby: > if iter will return a dict or a list: 

What should be returned by `__iter__` according to the documentation?

How `in` and `__iter__` work together?

E1-Bart0: `__in__`
`a __in__ (c, b, a):`
   `a __eq__ c`
   `a __eq__ b`
   `a __eq__ a`

`__iter__` in my case returns:
`(x, x, x)`
  `__in__` in my case:
  `a __eq__ x`
     `__eq__` in my case:
     `a in x.__dict__.values()`

`__iter__` if it will return a dict:
({'name': 'name1', 'age': 'age1'}, {'name': 'name2', 'age': 'age2'})
   `__in__` in my case:
  `name1 __eq__ {'name': 'name1', 'age': 'age1'} `
Or did I something miss?

E1-Bart0: >What should be returned by __iter__ according to the documentation?

> object implements iteration protocol. i.e. you could use it in for loops::
for president in presidents:
print(president['name'])

There is no info about what should returns in `__iter__` except it has `__getitem__`

Cjkjvfnby: > object implements iteration protocol. i.e. you could use it in for loops::

Not only iterators could be used in the for loops. 

> There is no info about what should returns in __iter__ 

Try to dig into https://docs.python.org

Cjkjvfnby: I am not sure if I understand you correctly
```
__eq__ in my case:
a in x.__dict__.values()
```


```
"a" == {1: "a", 2: "b"} # True
"b" == {1: "a", 2: "b"} # True
```

Than I hear equals I expect transitivity that means  "a" == "b" should be True as well;

All_home_works/hw8/task2.py
E1-Bart0: `        return Data(cursor=self.cursor)  # noqa: S608`
in this line is not actually needed. Are you talking about the previous line?
`        self.cursor.execute(f"SELECT * FROM '{self.db_table}'")  # noqa: S608`
I have looked at what injection in SQL means. And it seems we can't get rid of it in our case.

All_home_works/hw8/task2.py
E1-Bart0: Remade it

All_home_works/hw8/task2.py
E1-Bart0: `Select *` It would give us all fields at a table. (password, auth_token, e.c.)

All_home_works/hw8/task2.py
E1-Bart0: `__del__` will call when all references to the object have been deleted.
I think the better way to use it through `__open__` and `__exit__`, is it true?

test/hw8/test_task2.py
E1-Bart0: In general, I want to say that I do not know how to make a dictionary return in `for loop`, and when we do
`something IN storage`, it compared not with the dictionary, but with the values of the dictionary.
Because when we do it: `something IN storage` it do this: 
`for i in storage`
`if something == i`
`return True`

I mean it also calls `__iter__` and compares `something` with `__iter__` returns

All_home_works/hw8/task2.py
Cjkjvfnby: >  in this line is not actually needed. Are you talking about the previous line?

No, I have left comments precisely to this line.  It's a bad practice to add things that are not actually needed.

All_home_works/hw8/task2.py
Cjkjvfnby: What else could be bad here?  hint: `You know that thing`

All_home_works/hw8/task2.py
Cjkjvfnby: Yes

test/hw8/test_task2.py
Cjkjvfnby: I got your point. 

But lets me rewrite this code in different way: 

```python
i = next(storage)
while True:
    next(storage)
    if something == i
        return True
```

This is usly and won't work with most implementation of iterator protocall, except this.

All_home_works/hw8/task2.py
```diff
@@ -0,0 +1,46 @@
...
+        self.cursor = self._conn.cursor()
+        self.check_if_table_in_db()
+        return self
+
+    def __iter__(self):
+        self.cursor.execute(f"SELECT * FROM '{self._db_table}';")  # noqa: S608
+        return (dict(zip(data.keys(), data)) for data in self.cursor.fetchall())
+
+    def __contains__(self, item):
+        return any(item in dict_obj.values() for dict_obj in self)
```
Cjkjvfnby: What problem could we face here?

All_home_works/hw8/task2.py
```diff
@@ -0,0 +1,46 @@
...
+            raise sqlite3.OperationalError(f"no such table: {self._db_table}")
+
+    def __enter__(self):
+        self.cursor = self._conn.cursor()
+        self.check_if_table_in_db()
+        return self
+
+    def __iter__(self):
+        self.cursor.execute(f"SELECT * FROM '{self._db_table}';")  # noqa: S608
+        return (dict(zip(data.keys(), data)) for data in self.cursor.fetchall())
```
Cjkjvfnby: What problem could we face here?

All_home_works/hw8/task2.py
```diff
@@ -0,0 +1,46 @@
...
+        return self.cursor.fetchone()[0]
+
+    def __getitem__(self, item):
+        self.cursor.execute(
+            f"SELECT * FROM '{self._db_table}' where name=?;", (item,)  # noqa: S608
+        )
+        data = self.cursor.fetchone()
+        return dict(zip(data.keys(), data)) if data is not None else None
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
Cjkjvfnby, line 50: Do we always close the connection when we use this class?

All_home_works/hw8/task2.py
```diff
@@ -0,0 +1,46 @@
...
+    def __len__(self):
+        self.cursor.execute(f"SELECT count(*) FROM '{self._db_table}';")  # noqa: S608
+        return self.cursor.fetchone()[0]
+
+    def __getitem__(self, item):
+        self.cursor.execute(
+            f"SELECT * FROM '{self._db_table}' where name=?;", (item,)  # noqa: S608
+        )
+        data = self.cursor.fetchone()
+        return dict(zip(data.keys(), data)) if data is not None else None
```
Cjkjvfnby: This conversion looks ugly, can we make it better?

test/hw8/test_task2.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@pytest.fixture(scope="module")
+def create_db_table():
+    directory = tempfile.TemporaryDirectory()
+    path = os.path.join(directory.name, "test.sqlite")
+    db_name = os.path.normpath(path)
+    db_table = f"test_table_{time.time_ns()}"
+
+    conn = sqlite3.connect(db_name)
```
Cjkjvfnby: Actually, you could change db_name value and this will simplify your code a lot.

test/hw8/test_task2.py
```diff
@@ -0,0 +1,106 @@
...
+    with TableData(db_name, db_table) as storage:
+        assert 2 == len(storage)
+
+
+def test_table_data__if_len_correct__after_update_storage(updated_db):
+    storage = updated_db
+    assert 3 == len(storage)
+
+
+def test_table_data__get_row_from_db_by_name_if_not_such_row_none_expect(
```
Cjkjvfnby: Try to make this phrase sound more like a normal English phrase.

test/hw8/test_task2.py
```diff
@@ -0,0 +1,106 @@
...
+    create_db_table,
+):
+    db_name, db_table = create_db_table
+    with TableData(db_name, db_table) as storage:
+        assert {"name": "Book1", "author": "John Doe"} == storage["Book1"]
+        assert None is storage["Book3"]
```
Cjkjvfnby: Why is it None?

test/hw8/test_task2.py
```diff
@@ -0,0 +1,106 @@
...
+        assert "Jain Doe" in storage
+
+
+def test_table_data__if_something_in_storage__but_something_not_exists_in_storage(
+    create_db_table,
+):
+    db_name, db_table = create_db_table
+    with TableData(db_name, db_table) as storage:
+        result = "Book3" in storage
+        assert False is result
```
Cjkjvfnby:
```suggestion
        assert "Book3" not in storage
```

Cjkjvfnby (changes requested):

All_home_works/hw8/task2.py
E1-Bart0: - A lot of data.
- Something has changed in the data.

All_home_works/hw8/task2.py
E1-Bart0: I have opened the connection, not in `__enter__`. So the answer is: we do not always close the connection.

All_home_works/hw8/task2.py
E1-Bart0: Now I'm raising KeyError if there is no record.

All_home_works/hw8/task2.py
Cjkjvfnby: > A lot of data.

How we can handle it?

Cjkjvfnby (changes requested):
Please check the comments.

All_home_works/hw8/task2.py
E1-Bart0: We can each iteration take a new line from DB. But it's also not so fast as to take whole data. And we will have many requests to DB.

Cjkjvfnby: Let's do something in the middle.  Make a single request and read a row from DB on each iteration.

All_home_works/hw8/task2.py
```diff
@@ -26,24 +26,33 @@ def __enter__(self):
...
-        return (dict(zip(data.keys(), data)) for data in self.cursor.fetchall())
+        return self
+
+    def __next__(self):
+        data = self.cursor.fetchone()
+        if data is None:
+            raise StopIteration
+        return dict(zip(data.keys(), data))
 
     def __contains__(self, item):
```
Cjkjvfnby, line 31: What is the return type of this function?

Cjkjvfnby (changes requested):
Almost there.

All_home_works/hw8/task2.py
```diff
@@ -0,0 +1,58 @@
...
+    def __enter__(self):
+        self._conn = sqlite3.connect(self._db_name)
+        self._conn.row_factory = sqlite3.Row
+        self.cursor = self._conn.cursor()
+        self.check_if_table_in_db()
+        return self
+
+    def __iter__(self):
+        self.cursor.execute(f"SELECT * FROM '{self._db_table}';")  # noqa: S608
+        return self
```
Cjkjvfnby: Let's change this code a bit.   Instead of returning self let's return the iterator here.  After that, we can remove `__next__`

Cjkjvfnby (changes requested):

All_home_works/hw8/task2.py
```diff
@@ -26,13 +26,11 @@ def __enter__(self):
...
-        return self
+        return self.get_rows_from_cursor()
 
-    def __next__(self):
-        data = self.cursor.fetchone()
-        if data is None:
-            raise StopIteration
-        return dict(zip(data.keys(), data))
+    def get_rows_from_cursor(self):
+        while data := self.cursor.fetchone():
```
Cjkjvfnby: Almost there. What is the recomended way to iterate over rows?

Cjkjvfnby (changes requested):

Cjkjvfnby (approved):

PR #32, opened 2021-05-06
Hw9_task1

E1-Bart0:


All_home_works/hw9/task1.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import Iterator, List, Union
+
+
+def merge_sorted_files(file_list: List[Union[Path, str]]) -> Iterator:
+    num_generators = [_get_nums_from_file(file) for file in file_list]
+    numbers = [
+        (index, next(number, float("inf")))
+        for index, number in enumerate(num_generators)
+    ]
+    while not all(1 if num[1] == float("inf") else 0 for num in numbers):
```
Cjkjvfnby: `1 if num[1] == float("inf") else 0` could we make it simpler?

test/hw9/test_task1.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test__get_nums_from_file(filename):
+    expected = [1, 2, 3]
+    result = list(_get_nums_from_file(filename))
+    assert result == expected
+
+
+@patch("All_home_works.hw9.task1._get_nums_from_file", return_value=iter([1, 2, 3, 4]))
+def test__merge_sorted_files__check_one_file(_get_nums_from_file):  # noqa: PT019
```
Cjkjvfnby: What is PT019?

All_home_works/hw9/task1.py
Cjkjvfnby: Let's optimize it a bit for cases when we have many files, and files could have different lengths.

Cjkjvfnby (changes requested):

All_home_works/hw9/task1.py
Cjkjvfnby: Can you describe the result of optimization.   Let's try to talk on big O notation.   n -> number of files, k - the average number of numbers in the file.  

What was before, what you get after?

E1-Bart0: I understood my mistake and remade it

Cjkjvfnby: Let's optimize it a bit for cases when we have many files, and files could have different lengths.

Cjkjvfnby (changes requested):

All_home_works/hw9/task1.py
```diff
@@ -16,24 +16,27 @@
...
-
-
-def _get_min_number(numbers, num_generators, stop):
-    index, num = min(numbers, key=lambda x: x[1])
-    next_number = next(num_generators[index], float("inf"))
-    numbers[index] = (index, next_number)
-    if next_number == float("inf"):
-        stop += 1
-    return num, stop
+    numbers__generators = _get_numbers_and_generators(file_list)
```
Cjkjvfnby: It's more common to use a single underscore in names.  `snake_case` not `snake__case`.

Cjkjvfnby (approved):

PR #34, opened 2021-05-06
Hw9 task3

E1-Bart0:


Cjkjvfnby (changes requested):
This code does not meet the requirements.  Only for specific cases from the examples are supported.

E1-Bart0:
Could you explain part with tokenizer, maybe some significant examples, because requirements are terrible as and an example in it.

Cjkjvfnby:
> because requirements are terrible 

At least you have them.  In real life, you need to do a lot of digging yourself.  

In that case you violate the part that is described well.  Try to read requirements carefully.

E1-Bart0:
As I understand it:
Needed to count all lines in all files with such extension in dir. So all files mean all files even in subdir. I do not understand what I did wrong.

Cjkjvfnby:
> Needed to count all lines 

In the task description, it is written in a slightly different way.  Details matters.

E1-Bart0:
Oh. I got it. The thing is I cut `'\n'`.

E1-Bart0:
`It will count lines in all files with that extension`
So if the last line is empty, is It a line or not?

All_home_works/hw9/task3.py
```diff
@@ -0,0 +1,40 @@
...
+        counter += count_lines_in_file(file_path, tokenizer)
+    return counter
+
+
+def count_lines_in_file(file_path, tokenizer):
+    counter = 0
+    line = ""
+    with open(file_path) as file:
+        for line in file:
+            line = line
```
Cjkjvfnby: Nice

Cjkjvfnby (changes requested):
> So if the last line is empty, is It a line or not?

You called it `the last line` looks like it should be counted.

The code is still not doing what is required.  It's quite common mistake.  try to clear you mind and read requirements carefully.

All_home_works/hw9/task3.py
```diff
@@ -0,0 +1,44 @@
...
+def universal_file_counter(
+    dir_path: Path, file_extension: str, tokenizer: Optional[Callable] = None
+) -> int:
+    counter = 0
+    for file_path in get_files_from_directory(dir_path, file_extension):
+        counter += count_lines_in_file(file_path, tokenizer)
+    return counter
+
+
+def count_lines_in_file(file_path, tokenizer):
```
Cjkjvfnby: Does this name describe this function behavior well?

All_home_works/hw9/task3.py
```diff
@@ -0,0 +1,44 @@
...
+        if tokenizer is None:
+            return count_lines(file)
+        return len(tokenizer(file.read()))
+
+
+def count_lines(file):
+    counter = 0
+    for line in file:  # noqa: B007
+        counter += 1
+    counter += 1 if counter and line.endswith("\n") else 0
```
Cjkjvfnby, line 37: Let's add a bit more tests for that cases.

test/hw9/test_task3.py
```diff
@@ -0,0 +1,58 @@
...
+                dir=directory.name, suffix=extension, mode="w"
+            )
+            named_file.writelines(data)
+            named_file.seek(0)
+            files.append(named_file)
+        return directory.name
+
+    yield _dirname
+    for file in files:
+        file.close()
```
Cjkjvfnby: Can we close files using context manager?

test/hw9/test_task3.py
```diff
@@ -0,0 +1,58 @@
...
+    directory.cleanup()
+
+
+@pytest.mark.parametrize(("extension", "expected"), [("txt", 4), ("py", 2)])
+def test_universal_file_counter__without_tokenizer(make_dir, extension, expected):
+    data = ["one line\n", "second line"]
```
Cjkjvfnby, line 34: It looks strange that this code produces 4 items for txt and 2 for py.    Probably some magic is hidden in `make_dir`.

Could you rewrite this test in a way when it will be absolutely clear why you expect these numbers?

Cjkjvfnby (changes requested):

test/hw9/test_task3.py
E1-Bart0: If I remove these lines it will be an error: ` No such file or directory: '/tmp/...`.
I think it Is because in the actual function we are opening this file and then closing it. And while we are closing this file, the file deletes itself. And when the directory is cleaning up it is again trying to delete this not existing file, so I am catching the error.

All_home_works/hw9/task3.py
```diff
@@ -0,0 +1,45 @@
...
+    return counter
+
+
+def get_files_from_directory(dir_path, file_extension):  # noqa: CCR001
+    for dirpath, _, filenames in os.walk(dir_path):
+        for file in filenames:
+            if file.split(".")[-1] == file_extension:
```
Cjkjvfnby: Can we do it better?

All_home_works/hw9/task3.py
```diff
@@ -0,0 +1,45 @@
...
+def universal_file_counter(
+    dir_path: Path, file_extension: str, tokenizer: Optional[Callable] = None
+) -> int:
+    counter = 0
+    for file_path in get_files_from_directory(dir_path, file_extension):
+        with open(file_path) as file:
+            if tokenizer is None:
+                counter += count_lines(file)
+            else:
+                counter += count_tokens(file, tokenizer)
```
Cjkjvfnby: Can we simplify this code?

test/hw9/test_task3.py
Cjkjvfnby: This question is still actual.

test/hw9/test_task3.py
```diff
@@ -0,0 +1,66 @@
...
+def test_universal_file_counter__without_tokenizer_file_ends_with_empty_line(make_dir):
+    data = ["first line\n"]
+    dirname = make_dir(data, (".py",))
+    res = universal_file_counter(dirname, "py")
+    assert res == 2
+
+
+def test_universal_file_counter__with_tokenizer(make_dir):
+    expected = 4
+    data = ["one line\n", "second line\n"]
```
Cjkjvfnby: It is not clear, why are you expecting 4 items but has only 2 \n in code.

Cjkjvfnby (changes requested):

test/hw9/test_task3.py
E1-Bart0: But spaces are between words. If we count `\n` it is not clear what had happened. Like in the example.

All_home_works/hw9/task3.py
```diff
@@ -41,5 +42,5 @@ def count_lines(file):
 def get_files_from_directory(dir_path, file_extension):  # noqa: CCR001
     for dirpath, _, filenames in os.walk(dir_path):
         for file in filenames:
-            if file.split(".")[-1] == file_extension:
+            if file.endswith(f".{file_extension}"):
```
Cjkjvfnby: Let's find one more way to work with extensions.

All_home_works/hw9/task3.py
```diff
@@ -19,10 +19,11 @@ def universal_file_counter(
     counter = 0
     for file_path in get_files_from_directory(dir_path, file_extension):
         with open(file_path) as file:
-            if tokenizer is None:
-                counter += count_lines(file)
-            else:
-                counter += count_tokens(file, tokenizer)
+            counter += (
+                count_lines(file)
+                if tokenizer is None
```
Cjkjvfnby: If as called multiple times.  Let's call it once.

Cjkjvfnby (changes requested):

All_home_works/hw9/task3.py
E1-Bart0: `if pathlib.Path(file).suffix == f'.{file_extension}':`
`if os.path.splitext(file)[-1] == f'.{file_extension}':`

All_home_works/hw9/task3.py
```diff
@@ -12,26 +12,25 @@
...
+            counter += count_function(file, tokenizer)
     return counter
 
 
 def count_tokens(file, tokenizer):
     return len(tokenizer(file.read()))
 
 
-def count_lines(file):
+def count_lines(file, *args, **kwargs):
```
Cjkjvfnby: Why do we need args and kwargs?

Cjkjvfnby (changes requested):

All_home_works/hw9/task3.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def universal_file_counter(
+    dir_path: Path, file_extension: str, tokenizer: Optional[Callable] = None
+) -> int:
+    counter = 0
+    count_function = count_lines if tokenizer is None else count_tokens
+    for file_path in get_files_from_directory(dir_path, file_extension):
+        with open(file_path) as file:
+            counter += count_function(file, tokenizer)
```
E1-Bart0, line 25: Is it better to use `count_lines(file, tokenizer)` instead of `(*args, `**kwargs)` because it's more obvious?

Cjkjvfnby (approved):

PR #35, opened 2021-05-10
Hw10 task1

E1-Bart0:


Cjkjvfnby (changes requested):
What does `async`  mean when it is added to function?

E1-Bart0:
It will run asynchronously. If simplify it's mean when one function is do nothing, in this time will work another function

Cjkjvfnby:
> It will run asynchronously. If simplify it's mean when one function is do nothing, in this time will work another function

Kind of.  How do we know that this function does nothing?

E1-Bart0:
A function does not load a CPU

Cjkjvfnby:
> A function does not load a CPU

It is hard to measure.  Anything else in the function that could tell us about it?

E1-Bart0:
This needs some changes. I will re-request by the near future.

All_home_works/hw10/task1_get_result.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+    P.S. по-максимуму использовать возможности параллелизма и асинхронности.
+
+    bs4
+    aiohttp
+
+"""
+import asyncio
+from multiprocessing import Pool
```
Cjkjvfnby: What is the difference between async and multiprocessing?

All_home_works/hw10/task1_get_result.py
```diff
@@ -0,0 +1,140 @@
...
+            strip=True, separator=""
+        )
+        return float(course.replace(",", "."))
+
+    response = await fetch_response(COURSE_URL)
+    return _parse_current_course(response)
+
+
+async def collect_data():
+    global COURSE
```
Cjkjvfnby: Let's find a way to avoid using global.

test/hw10/test_task1_get_result.py
```diff
@@ -0,0 +1,227 @@
...
+    return_value=iter([("/company/1", 10.0), ("company/2", -10.0)]),
+)
+async def test_collect_data(get_companies_url, collect_html, get_course, get_response):
+    result = await collect_data()
+    get_response.assert_called_once_with("url")
+    get_course.assert_called_once()
+    get_companies_url.assert_called_once_with(get_response.return_value)
+    calls = collect_html.call_args_list
+    assert [call(("/company/1", 10.0)), call(("company/2", -10.0))] == calls
+    assert result == ["html", "html"]
```
Cjkjvfnby: it's better to use the name that describes the content: `company_page_content`

test/hw10/test_task1_get_result.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+@pytest.mark.asyncio()
+@patch("All_home_works.hw10.task1_get_result.fetch_response", return_value="text")
+@patch("All_home_works.hw10.task1_get_result.MAIN_URL", "url/")
+async def test_collect_html_for_companies(get_response):
+    data = ("company/1", 10.0)
+    result = await collect_html_for_companies(data)
+    get_response.assert_called_once_with("url/company/1")
+    assert result == ("text", 10.0)
```
Cjkjvfnby: You can skip parenthesis here.

test/hw10/test_task1_get_result.py
```diff
@@ -0,0 +1,227 @@
...
+    ("string", "expect"),
+    [
+        ("-11%", -11.0),
+        ("10.0%", 10.0),
+        ("1,000.00", 1000.0),
+        ("-1,000.00", -1000.0),
+    ],
+)
+def test_get_float(string, expect):
+    assert expect == get_float(string)
```
Cjkjvfnby: Actually, this is a good candidate for doctests.

test/hw10/test_task1_save_result.py
```diff
@@ -0,0 +1,134 @@
...
+    data = [
+        {"P/E": 3},
+        {"P/E": None},
+        {"P/E": 2},
+    ]
+    result = top_fewest_pe_rating_companies(data)
+    assert result == [{"P/E": 2}, {"P/E": 3}, {"P/E": None}]
+
+
+def test_top_most_growth_companies():
```
Cjkjvfnby, line 116: Actually, you don't test top_most, since you return all values. 

You could use optional arguments for testing:
top_most_growth_companies(data, limit=2)

All_home_works/hw10/task1_get_result.py
```diff
@@ -0,0 +1,140 @@
...
+    p_e_ratio = get_values(table, "P/E Ratio")
+    week_low = get_values(table, "52 Week Low")
+    week_high = get_values(table, "52 Week High")
+    if week_low is None or week_high is None:
+        return p_e_ratio, None
+    profit = round((week_high / week_low - 1) * 100, 2)
+    return p_e_ratio, profit
+
+
+def get_float(number: str):
```
Cjkjvfnby: `get` is a vague term.  Try to vary verbs where it is possible for example "parse_float"

All_home_works/hw10/task1_get_result.py
```diff
@@ -0,0 +1,140 @@
...
+COURSE = 1
+
+
+async def fetch_response(url):
+    async with aiohttp.ClientSession() as session:
+        async with session.get(url) as response:
+            return await response.text()
+
+
+async def collect_html_for_companies(data):
```
Cjkjvfnby: `collect_html_for_companies` - > `collect_html_for_company`

All_home_works/hw10/task1_get_result.py
```diff
@@ -0,0 +1,140 @@
...
+    company_url, annual_growth = data
```
Cjkjvfnby: Do you just pass annual_growth  and only return? 

```
(A, B) - > (foo(A), B)
```

All_home_works/hw10/task1_get_result.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+        return float(course.replace(",", "."))
+
+    response = await fetch_response(COURSE_URL)
+    return _parse_current_course(response)
+
+
+async def collect_data():
+    global COURSE
+    response, COURSE = await asyncio.gather(fetch_response(URL), get_current_course())
```
Cjkjvfnby: `response` name describes the technical side of variable: something that comes from the internet.   

I prefer to see that name that describes its purpose.    You get responses from currency, from the company list, and from the company page.   Since they all in the same project, it's better to give them unique and descriptive names.

Cjkjvfnby (changes requested):
I like this solution.   

Left some cosmetic comments.

All_home_works/hw10/task1_get_result.py
E1-Bart0: It's a huge difference. Asyncio works in one thread and can't work with a few functions at the same moment. Only one function works and when it starts sleeping or waiting, the control takes to another function.
And multiprocessing could work with few functions at the same moment in different processor cores, but they do not have a common memory.

All_home_works/hw10/task1_save_result.py
```diff
@@ -1,58 +1,58 @@
...
-        top_fewest_pe_rating_companies,
-        top_most_growth_companies,
-        top_most_profit_companies,
-    )
-    with Pool(os.cpu_count()) as pool:
-        pool.starmap(save_to_file, zip(func, repeat(data)))
+from All_home_works.hw10.task1_get_result import CollectDataAboutCompany
+
+
+class SaveDataAboutCompanies:
```
Cjkjvfnby: Why do you need a class here?

All_home_works/hw10/task1_get_result.py
```diff
@@ -43,98 +43,121 @@
...
-def get_float(number: str):
-    number = number.rstrip("%")
-    return float(number.replace(",", ""))
-
-
-def get_result():
-    data = asyncio.run(collect_data())
-    with Pool() as pool:
-        return pool.starmap(get_company_as_a_dict, data)
+class CollectDataAboutCompany:
```
Cjkjvfnby: The approach with the function was more clear and simple. 

To get rid of the global variables you just need to pass them as arguments.

Cjkjvfnby (changes requested):

E1-Bart0:
Andrey, hello=) I'm writing just to notify you that I need you to review.

All_home_works/hw10/task1_get_result.py
```diff
@@ -0,0 +1,165 @@
...
+
+"""
+import asyncio
+from itertools import repeat
+from multiprocessing import Pool
+
+import aiohttp
+from bs4 import BeautifulSoup
+
+MAIN_URL = "https://markets.businessinsider.com/"
```
Cjkjvfnby: Try to avoid constants:

```python
CONSTANT = ...

def foo():
   ...(CONSTANT)
```

This simple trick will save you some pathing during testing.


```python
def foo(value):
   ...(value)

def main():
   value = ...
   foo(value)
```

`main` is a setup of your application, all it does is put settings to your code.  A similar approach is used in the Django framework.  You have your app which reads settings from files and a couple of settings files (development settings, tests setting, production settings).   So in this example, the `main` analog of setting for running code on your machine and for tests we run code in a different way so we don't need to write tests for main.

Cjkjvfnby (approved):

E1-Bart0:
Доброго времени суток!
Огромное спасибо за уделенное время и помощь с домашними заданиями.
Также хочу поблагодарить за совет: "подключить линтеры на работе" =)
Спасибо за все!

PR #36, opened 2021-05-12
Hw6 task2

E1-Bart0:


All_home_works/hw6/task2.py
```diff
@@ -0,0 +1,119 @@
...
+            Teacher.homework_done = defaultdict(set)
+        else:
+            del Teacher.homework_done[homework]
+
+    @staticmethod
+    def create_homework(text, deadline):
+        return Homework(text=text, deadline=deadline)
+
+
+class InstanceNotHomeworkError(Exception):
```
Cjkjvfnby: For educational purposes, let's subclass it from a built-in Exception which suits this case.

All_home_works/hw6/task2.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class Human:
+    def __init__(self, first_name, last_name):
+        self.last_name = last_name
+        self.first_name = first_name
+
+
+class Student(Human):
+    def do_homework(self, homework, solution):
```
Cjkjvfnby, line 85: Let's remove some lines that are not needed here.

Cjkjvfnby (changes requested):
Have you implemented all requirements?

All_home_works/hw6/task2.py
E1-Bart0: Is it okay to check if homework is active in homework result or it will break logic?

Cjkjvfnby: > Is it okay to check if homework is active in homework result or it will break logic?

What is your opinion on this?

All_home_works/hw6/task2.py
```diff
@@ -106,14 +108,15 @@ def create_homework(text, deadline):
         return Homework(text=text, deadline=deadline)
 
 
-class InstanceNotHomeworkError(Exception):
+class InstanceNotHomeworkError(BaseException):
```
Cjkjvfnby: Why BaseException?

Cjkjvfnby (changes requested):

All_home_works/hw6/task2.py
E1-Bart0: > For educational purposes, let's subclass it from a built-in Exception which suits this case.

And BaseException is the base class for all built-in exceptions.

All_home_works/hw6/task2.py
E1-Bart0: I would prefer to check `isinstance` and `is_active` before making a result in `do_homework` and do not even initialize `hw_result` if it not correct. Because `hw_result` it's like a finished job and we shouldn't check `is_active` or `isinstance` in it.

All_home_works/hw6/task2.py
Cjkjvfnby: > And BaseException is the base class for all built-in exceptions.

What does documentation say about it?

All_home_works/hw6/task2.py
Cjkjvfnby: Looks good.

> I would prefer to check

Actually in a case when it's possible to reference requirements it's better to do it.  

`According to requirements` sounds more powerful.

Cjkjvfnby (changes requested):
Let's implement this. https://github.com/E1-Bart0/homework-repo/pull/36/files#r634083222

All_home_works/hw6/task2.py
E1-Bart0: > The base class for all built-in exceptions. It is not meant to be directly inherited by user-defined classes (for that, use Exception).

All_home_works/hw6/task2.py
```diff
@@ -101,20 +102,20 @@ def reset_results(homework=None):
...
 
     @staticmethod
     def create_homework(text, deadline):
         return Homework(text=text, deadline=deadline)
 
 
-class InstanceNotHomeworkError(BaseException):
+class InstanceNotHomeworkError(TypeError):
     def __init__(self, homework):
         self.messages = f"You gave a not Homework object, Got {type(homework)}"
```
Cjkjvfnby: `a not` -> `not a`


You can use some samples from the Python itself for error messages:  try `int([])`

Cjkjvfnby (approved):

All_home_works/hw6/task2.py
E1-Bart0: Like this?
`raise InstanceNotHomeworkError('You gave a not Homework object, Got', type(homework))`

Cjkjvfnby: I mean that your message could use a style that is used by Python error messages.

PR #38, opened 2021-05-15
Hw11 task2

E1-Bart0:


All_home_works/hw11/task2.py
```diff
@@ -0,0 +1,50 @@
...
+def elder_discount(order):
+    discount = 0.9
+    result = order.price - order.price * discount
+    return int(result)
+
+
+class Order:
+    def __init__(self, price, strategy):
+        self.price = price
+        self._strategy = strategy
```
Cjkjvfnby: What does name strategy mean in the context of Order?

All_home_works/hw11/task2.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @property
```
Cjkjvfnby: Why do we need it?

Cjkjvfnby (changes requested):

Cjkjvfnby (approved):

PR #42, opened 2021-05-23
Hw12

E1-Bart0:


requirements-dev.txt
```diff
@@ -1,3 +1,20 @@
+aiohttp==3.7.4.post0
```
Cjkjvfnby: Looks like you copy-pasted the output of the `pip freeze` command.

It is considered a bad practice. Usually, you specify only libs you are using directly like `alembic`.  

I'd recommend having 2 files with requirements for the project: one with requirements to run the project (alembic, etc) and the other for development staff (tests, linter, formatters).  So for development, you need to install from both of them and for production only from one of them.

All_home_works/hw12/migrations/script.py.mako
```diff
@@ -0,0 +1,24 @@
+"""${message}
```
Cjkjvfnby, line 1: Why do we need this file?

Cjkjvfnby (changes requested):

All_home_works/hw12/migrations/script.py.mako
E1-Bart0: This file is used to generate new migration scripts.

E1-Bart0:
Is it ok, or it needs some changes?

All_home_works/hw12/migrations/env.py
```diff
@@ -0,0 +1,83 @@
...
+    This configures the context with just a URL
+    and not an Engine, though an Engine is acceptable
+    here as well.  By skipping the Engine creation
+    we don't even need a DBAPI to be available.
+
+    Calls to context.execute() here emit the given string to the
+    script output.
+
+    """
+    # url = config.get_main_option("sqlalchemy.url")
```
Cjkjvfnby: ?

All_home_works/hw12/migrations/versions/ca54c2de187c_data_migration.py
```diff
@@ -0,0 +1,48 @@
...
+    user = Student(id=1, first_name="Jon", last_name="Doe")
+    teacher = Teacher(id=1, first_name="Jane", last_name="Doe")
+    homework1 = Homework(id=1, text="First", deadline=2)
+    homework_result = HomeworkResult(
+        id=1, solution="SOLUTION", author=user, homework=homework1, teacher_id=1
+    )
+    session.add_all([user, teacher, homework1, homework_result])
+
+    session.commit()
+    session.close()
```
Cjkjvfnby: Does session support `with`?

Cjkjvfnby (changes requested):
Look good. Some minor comments.

All_home_works/hw12/core/__init__.py
```diff
@@ -0,0 +1,9 @@
+from All_home_works.hw12.core.db import URL, Base, Session, engine, session_scope
+
+__all__ = [
```
Cjkjvfnby: This is required for asterisks import and asterisks import is not recommended. I prefer not to declare this variable at all.

Cjkjvfnby (approved):

